PR #20, opened 2022-09-27
Remove faulty reference to controller manager image in deploy

flawedmatrix:
It seems this file was preventing us from using our own controller image as specified by the IMG environment variable in the Makefile.


config/default/manager_image_patch.yaml
```diff
@@ -1,11 +0,0 @@
-apiVersion: apps/v1
-kind: Deployment
-metadata:
-  name: controller-manager
-  namespace: system
-spec:
-  template:
-    spec:
-      containers:
-      - image: gcr.io/cluster-api-ipam-provider-in-cluster/release/manager:main
```
schrej, line 10: Have you tried setting this to the correct image instead?

schrej (commented):
I didn't test deployments apart from Tilt so far, and also didn't adapt the things in config/ after adding automated image builds, sorry.

config/default/manager_image_patch.yaml
flawedmatrix: I haven't tried, but I believe it would work too. It just seemed like an extra unnecessary step considering that the Makefile uses the `IMG` environment variable.

schrej:
Thanks for the effort here.
I've chosen to keep the patch and adapted it to use the correct image. It's used in the release process now.

PR #45, opened 2022-11-23
Update pool to handle an ip set of free addresses

### LINKED ISSUE #21: Feature Request: Ability to configure the pool with a list of non-contiguous IP addresses
Feature Request:

We have a case where we're not able to obtain a contiguous block of IP addresses. We can, however, get a set of IPs that are on the same network. We'd like the ability to configure the pool with this list of IP. 

We propose being able to set a list of IP in place of a CIDR or start & end:

```
kind: InClusterIPPool
metadata:
  name: inclusterippool-sample
spec:
  prefix: 24
  gateway: 10.0.0.1
  addresses:
    - 10.0.0.5
    - 10.0.0.7
    - 10.0.0.9
    - 10.0.0.14
    - 10.0.0.6
```

It seems some validations would be prudent too, to ensure the IPs are within the gateway/prefix range.

cc @adobley @flawedmatrix @christianang

adobley:
To handle non-contiguous IPs within the pool, the FindFreeAddress function has been modified to accept an IPSet instead.

Validation requires when using Addresses that Gateway and Prefix are set. First, Last, and Subnet can not be set when using Addresses.

Closes #21


internal/poolutil/pool_test.go
```diff
@@ -0,0 +1,152 @@
...
+				})
+			})
+
+			It("returns a free address from that range", func() {
+				ip, err := FindFreeAddress(freeIPSet, existing)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(ip.String()).To(Equal("10.0.0.12"))
+			})
+		})
+
```
schrej, line 74: Since I wasn't sure that this is working while looking at the diff - could you add this test here?

```go
    When("only the last ip address is available", func() {
			BeforeEach(func() {
				existing = createIPSet([]iprange{
					{From: "10.0.0.10", To: "10.0.0.19"},
				})
			})

			It("returns that ip address", func() {
				ip, err := FindFreeAddress(freeIPSet, existing)
				Expect(err).NotTo(HaveOccurred())
				Expect(ip.String()).To(Equal("10.0.0.20"))
			})
		})
```

internal/poolutil/pool.go
```diff
@@ -63,18 +63,18 @@ func IPAddressListToSet(list []ipamv1.IPAddress, gateway string) (*netaddr.IPSet
...
+		ip := iprange.From()
+		for {
+			if !inUseIPSet.Contains(ip) {
 				return ip, nil
 			}
-			return netaddr.IP{}, errors.New("no address available")
+			if ip == iprange.To() {
+				break
+			}
+			ip = ip.Next()
```
schrej, line 76: Much cleaner this way, thanks!

internal/webhooks/inclusterippool.go
```diff
@@ -147,3 +156,50 @@ func (webhook *InClusterIPPool) validate(_, newPool *v1alpha1.InClusterIPPool) (
 
 	return //nolint:nakedret
 }
+
+func validateAddresses(newPool *v1alpha1.InClusterIPPool) field.ErrorList {
+	var allErrs field.ErrorList
+
+	if newPool.Spec.Subnet != "" {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "subnet"), newPool.Spec.Subnet, "subnet may not be used with addresses"))
```
schrej, line 164: not too sure, but it might be nice to allow using subnet. If we allow it, we should also default it to the correct subnet according to addresses and prefix.

internal/webhooks/inclusterippool.go
```diff
@@ -147,3 +156,50 @@ func (webhook *InClusterIPPool) validate(_, newPool *v1alpha1.InClusterIPPool) (
...
+	for i, address := range newPool.Spec.Addresses {
+		ip, err := netaddr.ParseIP(address)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "addresses"), address, "provided address is not a valid ip address"))
+			continue
+		}
+		ips[i] = ip
+	}
+
+	prefix, err := netaddr.ParseIPPrefix(fmt.Sprintf("%s/%d", newPool.Spec.Addresses[0], newPool.Spec.Prefix))
```
schrej, line 193: Subnet could be set to this prefix during defaulting.

schrej (changes requested):

schrej:
lgtm overall. Thanks for that large amount of tests. :raised_hands: 
Just a few smaller things and a few linter remarks.

internal/webhooks/inclusterippool.go
tylerschultz: We've responded to this in the previous comment.

api/v1alpha1/inclusterippool_types.go
```diff
@@ -6,8 +6,14 @@ import (
 
 // InClusterIPPoolSpec defines the desired state of InClusterIPPool.
 type InClusterIPPoolSpec struct {
+	// Addresses is an array of IP addresses that can  be assigned. This set of
```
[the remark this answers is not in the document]
tylerschultz: Will do.

internal/poolutil/pool_test.go
tylerschultz: Good eye!

internal/webhooks/inclusterippool.go
schrej: Valid point, let's keep it this way.

internal/poolutil/pool_test.go
schrej: I've trained it 😄 https://github.com/telekom/cluster-api-ipam-provider-in-cluster/commit/e0659c80f118a3ee360bb945eee2b7ccf019b4ec

schrej (approved):

schrej:
If you need a release that includes this, let me know.

tylerschultz:
Thanks @schrej! We've got our hands full with other priorities, so there's no urgency getting this feature released.

Issue #54, opened 2023-01-05
Proposal: Cross-Namespace Pools


adobley:
Currently an InClusterIPPool is per namespace. We would like to have the ability to define a single pool of IPs that can fulfill claims for clusters in various namespaces.

Right now it is possible to define the same pool in multiple namespaces and get the behavior of a shared pool. This has the overhead of maintaining that same resource in multiple namespaces. When changing the size of the pool all instances would need to be updated or risk configuration drift.

We propose adding a reference field to the `InClusterIPPool` resource called `poolPointer` which would be an [ObjectReference](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-reference/).

Using CAPV as an example users would create a template in the same way as before:
```
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
kind: VSphereMachineTemplate
metadata:
  name: example
  namespace: cluster-ns
spec:
  template:
    spec:
      cloneMode: FullClone
      numCPUs: 8
      memoryMiB: 8192
      diskGiB: 45
      network:
        devices:
        - dhcp4: false
          fromPool:
            group: ipam.cluster.x-k8s.io/v1alpha1
            kind: InClusterIPPool
            name: my-pool-pointer
```

Create the pool pointer:
```
apiVersion: ipam.cluster.x-k8s.io/v1alpha1
kind: InClusterIPPool
metadata:
  name: my-pool-pointer
  namespace: cluster-ns
spec:
  poolPointer:
    - name: my-pool
      namespace: pool-ns
      kind: InClusterIPPool
```

Create the pool:
```
apiVersion: ipam.cluster.x-k8s.io/v1alpha1
kind: InClusterIPPool
metadata:
  name: my-pool
  namespace: pool-ns
spec:
  pools:
    - subnet: 10.10.10.0/24
      start: 10.10.10.100
      end: 10.10.10.200
```

For pools that have a `poolPointer` the controller will need be updated to resolve the IPPool data from the provided namespace/name pool.

The webhook for `InClusterIPPool` should reject any resources that provide both `pools` and `poolPointer`.

`IPAddresses` are already created in the namespace of the `IPAddressClaim` and the controller currently lists `IPAddresses` from all namespaces when determining what IP to assign based on the `kind/name` of the pool.

We're happy to implement this and are looking for feedback on our proposed approach.

cc/ @tylerschultz @flawedmatrix @christianang

schrej:
[...]

PR #63, opened 2023-01-27
Cluster scoped pools

tylerschultz:
Introduces a GlobalInclusterIPPool that is scoped to the cluster.

IPAddressClaims from any namespace can reference pools of this type.

Fixes: #54


schrej (approved):
lgtm, thank you

This still contains the bug from the current main, which fetches IPAddresses from all namespaces for namespaced pools. You can fix it in here if you have the time, or I'll fix it in a subsequent PR.
I'd just add another method to the generic pool to fetch the namespace, and filter by it when fetching addresses when it's not empty.

tylerschultz:
Thanks @schrej. We'll create a separate PR to address the namespace bug, either today or tomorrow.

PR #86, opened 2023-03-08
Change logger from zap to klog

### LINKED ISSUE #69: Logs should include human readable timestamps
The current timestamps are integer format, which are helpful for log aggregators, but make it hard for humans to decipher. 

Proposal: Logs should be changed to ISO format, or should include both integer and ISO format timestamps.
```
2023-02-03T18:46:39Z
```

flawedmatrix:
- this keeps it consistent with other capi loggers such as CAPV
- this logger prints timestamps in a human readable format by default

Fixes #69


schrej (approved):
Thank you!

PR #88, opened 2023-03-14
Ensure existing IPAddress's finalizer and ownerRefs on pre-existing IPAddresses

tylerschultz:
Ensure IPAddresses have the expected ProtectAddress finalizer and Owner References to their respective InClusterIPPool and IPAddress Claims. A case where this may happen is when a cluster has been restored from backup.


internal/controllers/ipaddressclaim.go
```diff
@@ -163,26 +164,77 @@ func (r *IPAddressClaimReconciler) reconcile(ctx context.Context, c *ipamv1.IPAd
...
+	for i, ownerRef := range address.GetOwnerReferences() {
+		if ownerRef.APIVersion == poolGVK.GroupVersion().String() &&
+			ownerRef.Kind == poolGVK.Kind &&
+			ownerRef.Name == pool.GetName() {
+			poolRefIdx = i
+		}
+	}
+
+	address.OwnerReferences[poolRefIdx].Controller = pointer.Bool(false)
+	address.OwnerReferences[poolRefIdx].BlockOwnerDeletion = pointer.Bool(true)
```
srm09, line 225: What is this logic for?

internal/controllers/ipaddressclaim.go
```diff
@@ -163,26 +164,77 @@ func (r *IPAddressClaimReconciler) reconcile(ctx context.Context, c *ipamv1.IPAd
...
 		log.Info("Address is marked for deletion, but deletion is prevented until the claim is deleted as well.", "address", address.Name)
 	}
 
-	c.Status.AddressRef = corev1.LocalObjectReference{Name: address.Name}
+	claim.Status.AddressRef = corev1.LocalObjectReference{Name: address.Name}
 
 	return ctrl.Result{}, nil
 }
 
+func (r *IPAddressClaimReconciler) ensureOwnerRefsAndFinalizer(ctx context.Context, address *ipamv1.IPAddress, claim *ipamv1.IPAddressClaim, pool genericInClusterPool) error {
```
srm09, line 203: Should the logic be modified to check and update only if necessary? Also, maybe using a `PatchHelper` from the [CAPI utils](https://github.com/kubernetes-sigs/cluster-api/blob/323e257147a9a9094dde1215aada4e33a0788278/util/patch/patch.go#L52) be useful here.
You can find instances of this usage in the CAPI/CAPV codebases

srm09 (commented):
Some suggestions

internal/controllers/ipaddressclaim.go
tylerschultz: The helper functions that ensure the owner ref do not expose setting the Controller nor the BlockOwnerDeletion flag. The logic that creates the owner refs when the IPAddress is created sets these flags. We add those flags for consistency.
https://github.com/telekom/cluster-api-ipam-provider-in-cluster/blob/f67d96e92729d54fa51626d92d8adbb2d345527a/pkg/ipamutil/address.go#L27-L28

internal/controllers/ipaddressclaim.go
tylerschultz: Updated.

We didn't realize the patch helper would avoid the API call if no mods were found. Today we learned.

tylerschultz:
Hi @srm09, We made the refactor we discussed over a call.

We stopped creating the IPAddress with ownerRefs, and instead made a call to the ensure function to add the ownerRefs. There is only one way that ownerRefs and finalizers are added to an address. 

We didn't go as with the refactor as you suggested. We didn't want to create an address, and then patch it right away. We wanted to avoid a second API call when creating an address. This is confusing to describe, let us know if you'd like to discuss further. Cheers.

internal/controllers/ipaddressclaim.go
srm09: Got the answer for this one.

internal/controllers/ipaddressclaim.go
```diff
@@ -182,6 +198,68 @@ func (r *IPAddressClaimReconciler) reconcile(ctx context.Context, claim *ipamv1.
...
+		return errors.Wrap(err, "failed to patch address")
+	}
+	return nil
+}
+
+func (r *IPAddressClaimReconciler) createAddress(ctx context.Context, claim *ipamv1.IPAddressClaim, pool genericInClusterPool, addressesInUse []ipamv1.IPAddress) (*ipamv1.IPAddress, error) {
+	address, err := r.allocateAddress(ctx, claim, pool, addressesInUse)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to allocate address")
+	}
```
srm09, line 221: `allocateAddress` still does the job of setting the owner Refs. The call to `address := ipamutil.NewIPAddress(claim, pool)` in the function still sets the owner reference on the `IPAddress` object.

Ideally,
1. Fetch the IP address for the claim
2. If not found,  `allocateAddress` => **just** allocates a new unused IP address (does not make any API calls to persist the object, no owner ref logic in there either)
3. Use this IP address from above to set/reset finalizer and owner refs (making it clear that the controller owner ref is being set in the controller reconcile loop)
4. Persist the object to the API server.

This logic works for either scenarios (new/existing address). All the fanciful owner ref and finalizer reconciliation logic stays in one place. 
For the purpose of logging the presence of an existing IPAddress object, add an info logger after the fetch call for the IPAddress, since that was the main concern around the refactor.

tylerschultz: The patch helper is not able to create an address for the create case. I think there's something I'm missing.

srm09: I get your point now, there is a way around it https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrPatch
But am gonna stop pushing for this one now. The code does what it is supposed to, just maintaining it will be difficult in the longer term. There is an easy chance that we might forget updating either of the places in the future.

internal/controllers/ipaddressclaim_test.go
```diff
@@ -852,6 +853,124 @@ var _ = Describe("IPAddressClaimReconciler", func() {
...
+					Name:      "test",
+					Namespace: namespace,
+					OwnerReferences: []metav1.OwnerReference{
+						{
+							APIVersion: "alpha-dummy",
+							Kind:       "dummy-kind",
+							Name:       "dummy-name",
+							UID:        "abc-dummy-123",
+						},
+					},
```
srm09, line 922: Consider adding 2 cases:
1. no owner references
2. unrelated owner refs (which is already present)

srm09 (commented):
minor suggestion, but overall lgtm

internal/controllers/ipaddressclaim.go
schrej: `ipamutil.NewIPAddress` was changed to no longer set the owner references. Maybe you've missed that change below?

I like it this way.

We could consider moving `ensureOwnerRefs` (without the finalizer part) to `ipamutil` as well. The idea is to potentially move `ipamutils` to CAPI to allow reusing some logic between providers. We might be able to improve it further with a second provider (e.g. the infoblox one) and turn it into some kind of frameworks. There should be quite a few similarities between ipam providers I think.

tylerschultz: @srm09 Ah, we got fixated on using the patch helper code, I didn't realize there was a CreateOrPatch. We've updated the logic as you've described. The result looks nice.

@schrej We've moved the ensureAddressOwnerReferences function to the `pkg/ipam-util` package.

I think we've updated everything to everyone's preference. Let us know if we've missed something.

internal/controllers/ipaddressclaim_test.go
tylerschultz: Added test, resolving.

schrej (approved):
Looks very clean now, thanks!

@tylerschultz could you fix the linter issue? It's a new problem now, I've fixed the gci import order thing that it was complaining about before.

Issue #89, opened 2023-03-14
Pools should allow for NOT specifying gateways


tylerschultz:
The IPAM CRDs supplied by CAPI suggest that Gateway is not a required field. CAPV is written so that gateway can be supplied on the VSphereMachine, removing the need to configure it on the pool. CAIP Validation allows for an empty Gateway, but later fails to Parse the empty Gateway here:

https://github.com/telekom/cluster-api-ipam-provider-in-cluster/blob/main/internal/poolutil/pool.go#L59

CAIP should honor it's validations and not error on an empty gateway.

tylerschultz:
This issue is a duplicate of this other I opened a while back. 
https://github.com/telekom/cluster-api-ipam-provider-in-cluster/issues/70
--Sir-spam-a-lot

PR #90, opened 2023-03-14
Gateway can be omitted from pool spec

tylerschultz:
This change allows the Gateway to be omitted from the Pool, preventing an error. Also, prior to this change, Gateway was required when specifying `pool.spec.addresses`. This change makes gateway optional when specifying `pool.spec.addresses`.

We took this opportunity to move what we considered business logic out of the pool utility package, and brought it back closer to the other business logic in the controller. The gateway is conditionally added to the list of in use IPs in the controller.

fixes: #89


schrej:
Sorry, might have been the wrong order to merge these... Could you rebase this one please?

tylerschultz:
We expected a conflict, NP. Rebased, should be good now.

PR #93, opened 2023-03-21
Fix error in example in README.md

nickperry:
Changed first / last to start / end.

Fixes https://github.com/telekom/cluster-api-ipam-provider-in-cluster/issues/92


nickperry:
Closing as it looks like it was intentional to move from start / end to first / last, so the CRD and readme are correct and there was just one stale example that had lead me to believe start / end were the legitimate names for v1alpha1.

Issue #71, opened 2023-02-03
Pools fail to render the first/last - start/end IP addresses in the kubernetes cli


tylerschultz:
Pools fail to show the first/last or start/end values of pools.
There is also some mixing of first/last and start/end. Perhaps one set should be eliminated.

```bash
$ cat pool.yaml
---
apiVersion: ipam.cluster.x-k8s.io/v1alpha1
kind: InClusterIPPool
metadata:
  name: inclusterippool-sample
  namespace: default
spec:
  subnet: 10.0.0.0/24
  gateway: 10.0.0.1
  start: 10.0.0.22
  end: 10.0.0.40
$ kubectl apply -f pool.yaml
inclusterippool.ipam.cluster.x-k8s.io/inclusterippool-sample created
$ kubectl get inclusterippool
NAME                     SUBNET        FIRST   LAST
inclusterippool-sample   10.0.0.0/24
```

schrej:
No. I actually forgot to update the json tag in the struct. So on the API it's now called start/end, and in code it's first/last. We should choose one variant.
Any preference? I'd go for first/last.
I think we can justify breaking api changes since there is no proper release yet. Any objections?

tylerschultz:
We'd prefer to avoid breaking changes, if possible. We like first/last too.

OK, can o worms time: What if we expanded the Addresses fields so that they could accept ranges, something like this:

```yaml
---
apiVersion: ipam.cluster.x-k8s.io/v1alpha1
kind: InClusterIPPool
metadata:
  name: pool-name
  namespace: pool-ns
spec:
  subnet: 10.10.0.0/16
  gateway: 10.10.10.1
  addresses:
  - "10.10.9.200-10.10.9.210"
  - 10.10.9.219
```

This allows users to carve up their subnets in more complex ways, and we could also deprecate/remove first/last start/end.  We can also open another issue to discuss that proposal separately if you'd prefer.

schrej:
Good idea. Let's do it like that then.
Is kubectl able to properly display lists as printcolumns?

tylerschultz:
Ah, good point, I'm not sure. We'll see what we can come up with.

tylerschultz:
EndpointSlices print addresses in a comma separated list. It looks like by default, kubectl will bracketize a list, space separated.

PR #95, opened 2023-03-21
Addresses can be specified as hyphenated ranges and CIDRs

tylerschultz:
This PR makes it possible to declare an InClusterIPPool like the following example:

```yaml
kind: InClusterIPPool
metadata:
  name: inclusterippool-sample
spec:
  addresses:
    - 10.0.0.10
    - 10.0.0.24-10.0.0.32
    - 10.0.0.128/28
  prefix: 24
  gateway: 10.0.0.1
```

This PR also adds the addresses printcolumn to the InClusterIPPool resource, like this example:
```bash
$ kubectl get inclusterippool
NAME                     SUBNET   FIRST   LAST   ADDRESSES
inclusterippool-sample                           ["10.0.0.10-10.0.0.11","10.0.0.30/30","10.0.0.22"] 
```

This PR does not address the missing subnet nor First and Last columns.

This PR adds a test to ensure the controller allocates the lowest available IP in a given IP Pool.

This PR works towards some of the topics in Issue #71


schrej (approved):
lgtm!

PR #101, opened 2023-03-29
Generic ClaimReconciler

schrej:
This is an attempt to implement a generic `ClaimReconciler` that can be used by other IPAM providers as well. Most of the logic will probably be similar, with only the actual ip address allocation being provider specific. There are several benefits to this approach imo:

- When the ClaimReconciler is well tested, providers only need to implement specific cases (e.g. special conditions)
- It's a lot easier to ensure providers adhere to the IPAM contract. If everyone is using it, it would also be easier to change the contract.
- The ClaimHandler can be tested individually.

I'll validate the approach with the implementation of the infoblox provider I'm working on.

Ultimately this _could_ be moved to upstream CAPI.

@tylerschultz, @srm09 et. al - it would be great if you can have a look.

### DIFF
--- a/pkg/ipamutil/reconciler.go
+++ b/pkg/ipamutil/reconciler.go
@@ -1,0 +1,191 @@
+package ipamutil
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	ipamv1 "sigs.k8s.io/cluster-api/exp/ipam/api/v1alpha1"
+	clusterutil "sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/cluster-api/util/annotations"
+	"sigs.k8s.io/cluster-api/util/patch"
+	"sigs.k8s.io/cluster-api/util/predicates"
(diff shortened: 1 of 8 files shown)


pkg/ipamutil/reconciler.go
```diff
@@ -0,0 +1,175 @@
...
+	client.Client
+	Scheme *runtime.Scheme
+
+	WatchFilterValue string
+
+	Provider ProviderIntegration
+}
+
+// ProviderIntegration is an interface that must be implemented by an IPAM provider.
+type ProviderIntegration interface {
```
schrej, line 39: Not 100% happy with the name here.

pkg/ipamutil/reconciler.go
```diff
@@ -0,0 +1,175 @@
...
+const (
+	// ReleaseAddressFinalizer is used to release an IP address before cleaning up the claim.
+	ReleaseAddressFinalizer = "ipam.cluster.x-k8s.io/ReleaseAddress"
+
+	// ProtectAddressFinalizer is used to prevent deletion of an IPAddress object while its claim is not deleted.
+	ProtectAddressFinalizer = "ipam.cluster.x-k8s.io/ProtectAddress"
+)
+
+// ClaimReconciler reconciles a IPAddressClaim object using a ProviderIntegration.
+type ClaimReconciler struct {
```
schrej, line 29: Should we have a function to initialise this?

schrej:
/close
in favor of https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/pull/203

PR #109, opened 2023-04-04
support pausing pools

adobley:
InClusterIPPools support a .spec.paused property which will pause the reconciliation of claims referencing that pool

### DIFF
--- a/internal/controllers/ipaddressclaim.go
+++ b/internal/controllers/ipaddressclaim.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	ipamv1 "sigs.k8s.io/cluster-api/exp/ipam/api/v1alpha1"
+	"sigs.k8s.io/cluster-api/util/annotations"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/cluster-api/util/predicates"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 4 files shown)


schrej:
We already have support for the [paused annotation](https://github.com/telekom/cluster-api-ipam-provider-in-cluster/blob/main/internal/controllers/ipaddressclaim.go#L72).
Is that sufficient or would you prefer to have the field as well?

adobley:
It looked to us that this annotation is set on the `IPAddressClaim` and that there is not currently a mechanism to pause a Pool resource.

We thought it would be useful to allow operators to pause at a pool level.

Does it make sense here that we're exposing this as a spec property rather than using the annotation? This is more aligned with what we've seen in the Cluster spec, but then there is a diff in how to pause claims vs pools that I'm not happy with.

schrej:
Oh, sorry, I went over this too quickly.
Definitely makes sense to pause on pool level as well.

Also not sure what's better here either. Properties are definitely more user-friendly. But at the same time annotations are easier to handle with generic tools.
Could you check if there is a plan for harmonising this on CAPI? ClusterClass doesn't have a paused field for example.

christianang:
> Also not sure what's better here either. Properties are definitely more user-friendly. But at the same time annotations are easier to handle with generic tools.
> Could you check if there is a plan for harmonising this on CAPI? ClusterClass doesn't have a paused field for example.

My 2c after talking with @tylerschultz. I think it may make sense for us to continue the pattern of using the annotation. I think what swayed us was the fact that every resource other than cluster just relies on the annotation and it lets us utilize the generic tooling that exists. I'll work with @adobley to update the PR to use annotations.

christianang:
Updated the PR to use the pause annotation.

internal/index/index.go
```diff
@@ -14,25 +14,44 @@ import (
...
+		panic(fmt.Sprintf("Expected an IPAddressClaim but got a %T", o))
+	}
+	return []string{IPPoolRefValue(ip.Spec.PoolRef)}
 }
 
-// IPAddressPoolRefValue turns a corev1.TypedLocalObjectReference to an indexable value.
-func IPAddressPoolRefValue(ref corev1.TypedLocalObjectReference) string {
+// IPPoolRefValue turns a corev1.TypedLocalObjectReference to an indexable value.
+func IPPoolRefValue(ref corev1.TypedLocalObjectReference) string {
 	return fmt.Sprintf("%s%s", ref.Kind, ref.Name)
```
schrej, line 56: Should we include namespace here to properly handle pools with the same name in different namespaces? Probably just an efficiency thing though.

internal/controllers/ipaddressclaim.go
```diff
@@ -126,6 +169,11 @@ func (r *IPAddressClaimReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		pool = gicippool
 	}
 
+	if annotations.HasPaused(pool) {
+		log.Info("IPAddressClaim references Pool which is paused", "IPAddressClaim", claim.GetName(), "Pool", pool.GetName())
```
schrej, line 173: Should we indicate that we're skipping reconciliation?

schrej (approved):
two nits, otherwise lgtm

internal/index/index.go
christianang: Yeah it makes sense to reduce the possibility to reconcile claims that have the same pool name in different namespaces. We decided to go with the approach of using `client.InNamespace` option when calling list which should also ensure the list call only looks in the namespace of the pool.

Note: We did also try it with GlobalInClusterIPPool, which doesn't have a namespace and it does seem to be fine with empty namespace.

PR #113, opened 2023-04-06
add explicit securityContexts to the controller

chrischdi:

This does not really change the configuration, it just makes it explicit and enforce the defaults, except for the seccompPolicy which changes from Unconfined to RuntimeDefault. Syscalls filtered by RuntimeDefault policy are 95% namespaced and require capabilities (which we drop) in the first place, so no practical change there either.

This allows to be compatible to the restricted pod security admission profile.

This is a recommendation in the [CAPI v1.3->1.4 upgrade guide](https://main.cluster-api.sigs.k8s.io/developer/providers/migrations/v1.3-to-v1.4.html#suggested-changes-for-providers).

**Special notes for your reviewer**:

Prior art:
* https://github.com/kubernetes-sigs/cluster-api/pull/7831
* https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4104

Note: this requires the v1.4 branch of CAPI for usage with tilt, because the Tiltfile then removes the securityContext for allowing it to succeed.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -31,8 +31,6 @@ spec:
         - --leader-elect
         image: controller:latest
         name: manager
-        securityContext:
-          allowPrivilegeEscalation: false
         livenessProbe:
           httpGet:
             path: /healthz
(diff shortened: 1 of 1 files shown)


schrej (approved):
Great, thank you!

PR #116, opened 2023-04-11
Pools show usage counts

### LINKED ISSUE #112: Pool usage metrics
It would be useful if we could some usage statistics from the pool without having to query the IPAddresses and count ourselves.

We are thinking of adding a total, used, free in the pool status. E.g.

```yaml
---
kind: InClusterIPPool
metadata:
  name: pool-name
  namespace: pool-ns
spec:
  subnet: 10.0.0.0/24
status:
  ipAddresses:
    # The total number of IPs the pool has.
    total: 255
    # The number of IPs that have already been allocated.
    used: 9
    # The number of IPs that are available to allocate.
    free: 250
```

tylerschultz:
InClusterIPPool and GlobalInClusterIPPool receive updates to their status showing the Total count of IPs in the pool's range, the InUse count, and the Free count.

IPv6 IP ranges can be enormous (2^128) IPs. Range totals that exceed int are shown as the value of math.MaxInt (arch dependent).

Adds a new reconciler for pools.
Adds columns to pools for Total, Free, Used

Remove Zap logger remnants.

Fixes #112


api/v1alpha1/inclusterippool_types.go
```diff
@@ -41,6 +41,27 @@ type InClusterIPPoolSpec struct {
 
 // InClusterIPPoolStatus defines the observed state of InClusterIPPool.
 type InClusterIPPoolStatus struct {
+	// IPAddresses reports the count of total, free, and used IPs in the pool.
+	// +optional
+	IPAddresses InClusterIPPoolStatusIPAddresses `json:"ipAddresses,omitempty"`
```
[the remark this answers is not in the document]
tylerschultz: We like the suggestion. We've made the rename.

tylerschultz:
We've rebased, and resolved test failures that resulted. 

We see that the new controller causes Claims to be reconciled. The scenario where a claim is paused needed special handling. The 'filter' for claims that are not paused are somehow bypassed by the changes the new controller makes to the pool. We've added a new pause check on the claim, which seems correct in any case.

We'll spend some time this afternoon manually testing the pause functionality along with these new usage counts and report back later.

schrej:
Since the IPAddressClaim controller has a watch on the pools, and requeues all claims for reconciliation whenever they are updated, this change now causes all claims to be reconciled whenever a new claim is created.
new claim -> status update on pool -> all claims reconciled.
Can we change the watches on the IPAddressClaim controller so it only queues when there is a change to the spec to avoid unnecessary reconciliations?
We can do that in a separate PR though.

christianang:
Yeah, that makes sense. Seems simple enough to just do in this PR. We will make an update.

tylerschultz:
That was a great catch @schrej. We've updated the PR to prevent all of the pool's claims from reconciling when the pool status is updated.

internal/controllers/ipaddressclaim.go
```diff
@@ -358,3 +357,19 @@ func buildAddressList(addressesInUse []ipamv1.IPAddress, gateway string) []strin
 
 	return addrStrings
 }
+
+func resourceTransitionedToUnpaused() predicate.Predicate {
+	return predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			_, wasPaused := e.ObjectOld.GetAnnotations()[clusterv1.PausedAnnotation]
+			_, nowPaused := e.ObjectNew.GetAnnotations()[clusterv1.PausedAnnotation]
```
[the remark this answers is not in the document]
schrej: could even be
```go
return annotations.HasPaused(e.ObjectOld) && !annotations.HasPaused(e.ObjectNew)
```

api/v1alpha1/inclusterippool_types.go
```diff
@@ -41,6 +41,27 @@ type InClusterIPPoolSpec struct {
...
+
+	// Free is the count of unallocated IPs in the pool.
+	// Counts greater than int can contain will report as math.MaxInt.
+	// +optional
+	Free int `json:"free,omitempty"`
+
+	// Used is the count of allocated IPs in the pool.
+	// Counts greater than int can contain will report as math.MaxInt.
+	// +optional
+	Used int `json:"used,omitempty"`
```
schrej, line 64: maybe we should remove `omitempty` for all of these. Otherwise `0` will not be shown, and `Free: 0` would be pretty useful.

schrej (commented):
lgtm with a few nits

api/v1alpha1/inclusterippool_types.go
tylerschultz: Ya, we tried that. Unfortunately, 0 is considered empty. Unless you know of a way to get around that, I think we're stuck with omitempty.

tylerschultz: Thinking more about it, what I said is probably technically not true - it's the optional setting that rejects 0. We did try different combinations of optional and omitempty and ended back with where we are.

internal/controllers/ipaddressclaim.go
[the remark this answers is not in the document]
tylerschultz: nice! done.

api/v1alpha1/inclusterippool_types.go
schrej: There is one way to make this work: turning Addresses into a pointer. I generally try to avoid pointers, but maybe it's worth it for the UX here? On the other hand, not being able to see the 0 is not that big of a deal.

tylerschultz: Of course, yes. Done!

api/v1alpha1/inclusterippool_types.go
```diff
@@ -41,6 +41,27 @@ type InClusterIPPoolSpec struct {
 
 // InClusterIPPoolStatus defines the observed state of InClusterIPPool.
 type InClusterIPPoolStatus struct {
+	// Addresses reports the count of total, free, and used IPs in the pool.
+	// +optional
+	Addresses InClusterIPPoolStatusIPAddresses `json:"ipAddresses,omitempty"`
```
schrej, line 46: I meant turning this into a pointer, and removing optional and omitempty from the ints in the struct. Should've written that more clearly before, sorry.

schrej (changes requested):

api/v1alpha1/inclusterippool_types.go
tylerschultz: Ah, yes, ok. I've made the status a pointer and de-pointerized the counts. All seems seems good.

tylerschultz:
Hi Jakob, we think the PR is once again ready for another round of review.

schrej (approved):

schrej:
Thank you!

Issue #70, opened 2023-02-03
Gateway should be optional


tylerschultz:
As it is written now, CAPV assumes that a Gateway will be present on `IPAddress` objects, and will fail to reconcile a `VSphereVM` if it's not present.
The webhook should validate the presence of the Gateway, and reject pools that do not have a valid gateway. Gateways should need to be in range of the provided `subnet`.
Alternatively, the gateway could be defaulted to the first IP in the subnet range.

schrej:
I gave this some more thought. If you want to rely on a Gateway being set, this would need to be specified on IPAM contract level, not for individual providers.

The fun part starts once you consider IPv6. (Disclaimer: I'm not too deep into IPv6, so feel free to validate)
Afaik there is a lot of new ways to configure IP addresses. There are better alternatives to DHCP, which makes the IPAM contract unnecessary in some scenarios. But there are also options to configure individual aspects of the network using "Router Advertising", which e.g. allows setting a static IPv6 on the host, but not providing a default gateway, which will be configured automatically using Router Advertising.
Considering that, I'd keep this field optional. Alternatively we could only make it required when using IPv4. I'll ask my networking colleagues about that as well, maybe there is a reason not to have it for IPv4 either.

Of course nobody thought of us poor software engineers who only want to understand the manageable complexity of IPv4 when coming up with those ideas. :smile:

I'm not sure what a solution for different requirements of providers would be. Of course it should be documented and be properly reflected in resource status. We could also introduce a toggle that makes it required, which would help folks with a less-advanced networking setup.

tylerschultz:
Thanks for bringing up these points. We agree that there is valid reason to keep it optional.

We do see that right now, the code will fail reconciliation if an invalid gateway is provided:
https://github.com/telekom/cluster-api-ipam-provider-in-cluster/blob/main/internal/poolutil/pool.go#L59

Perhaps for the time being, we guard against an invalid (empty) gateway, and defer making any decision about validation?

christianang:
I noticed that the capi-webhook validates gateway and it fails if gateway is empty. Probably have to update the capi-webhook to allow empty gateway to truly allow IPAddresses to get reconciled with empty gateways: https://github.com/kubernetes-sigs/cluster-api/blob/4f60841913fa99912771c3906ee81dd13d4e9633/exp/ipam/internal/webhooks/ipaddress.go#L126-L134

schrej:
[...]

PR #118, opened 2023-04-13
Add validation for IP family and gateways

tylerschultz:
- require IPv4 Gateway if addresses are IPv4
- continue to not require IPv6 Gateway if addresses are IPv6
- validate addresses are of the same family i.e cannot mix IPv4 and IPv6 addresses in a single pool
- validate addresses and gateway are of the same family (if gateway is provided)

Relates to #70

We did some experimenting around cases with and without Gateway. We think this PR represents common understandings and preferences. This PR puts forth that Gateway should be required with pools/addresses that are IPv4, and Gateway should be optional when addresses are IPv6. We're open to feedback if these choices seem incorrect or too heavy handed.

We see that (at least in our envs) that we are able to launch IPv6 machines without configuring a Gateway. The default route is provided by Router Advertisements. This leads us to the conclusion that there are valid cases when using IPv6 that Gateway should not be required.

Asides:
We intend to PR CAPI so that Gateway is optional when the IPAddress contains an IPv6 address. We intend to PR CAPI so that it requires Gateway when the address is IPv4.

We intend to PR CAPV so that Gateway is optional when the IPAddress is IPv6.

### DIFF
--- a/internal/webhooks/inclusterippool.go
+++ b/internal/webhooks/inclusterippool.go
@@ -191,17 +191,29 @@ func validateAddresses(newPool types.GenericInClusterPool) field.ErrorList {
 		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "prefix"), newPool.PoolSpec().Prefix, "a valid prefix is required when using addresses"))
 	}
 
+	var hasIPv4Addr, hasIPv6Addr bool
+	for _, address := range newPool.PoolSpec().Addresses {
+		ipSet, err := poolutil.AddressToIPSet(address)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "addresses"), address, "provided address is not a valid IP, range, nor CIDR"))
+			continue
+		}
+		from := ipSet.Ranges()[0].From()
+		hasIPv4Addr = hasIPv4Addr || from.Is4()
+		hasIPv6Addr = hasIPv6Addr || from.Is6()
+	}
+	if hasIPv4Addr && hasIPv6Addr {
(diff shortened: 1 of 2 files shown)


schrej (approved):
lgtm. let's wait for the capi decision though.

tylerschultz:
Given the [decision made in CAPI](https://github.com/kubernetes-sigs/cluster-api/issues/8536), this PR has been rebased and updated to leave Gateway optional for both IPv4 and IPv6.

internal/webhooks/inclusterippool.go
```diff
@@ -191,18 +191,36 @@ func validateAddresses(newPool types.GenericInClusterPool) field.ErrorList {
 		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "prefix"), newPool.PoolSpec().Prefix, "a valid prefix is required when using addresses"))
 	}
 
-	if newPool.PoolSpec().Gateway != "" {
-		_, err := netip.ParseAddr(newPool.PoolSpec().Gateway)
-		if err != nil {
-			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "gateway"), newPool.PoolSpec().Gateway, err.Error()))
+	for _, address := range newPool.PoolSpec().Addresses {
+		if !poolutil.AddressStrParses(address) {
```
schrej, line 195: Tbf this is just a performance nit, but can we get rid of `AddressStrParses` and just check for the error in the second loop below? Currently we're parsing each entry twice, that's not really necessary.

schrej (changes requested):

internal/webhooks/inclusterippool.go
tylerschultz: That makes sense, updated.

schrej (approved):

PR #122, opened 2023-04-18
🐛 Add cluster to scheme and rbac

flawedmatrix:
Unsure how we missed this, but we are seeing the following error from the controller
```
E0417 17:55:21.003315       1 logr.go:279] "controller-runtime/source: kind must be registered to the Scheme" err="no kind is registered for the type v1beta1.Cluster in scheme \"pkg/runtime/scheme.go:100\""
```

This PR addresses this bug by adding Cluster to the scheme and RBAC role.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -7,6 +7,14 @@ metadata:
   name: manager-role
 rules:
 - apiGroups:
+  - cluster.x-k8s.io
+  resources:
+  - clusters
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - ipam.cluster.x-k8s.io
   resources:
   - globalinclusterippools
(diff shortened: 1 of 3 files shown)


srm09:
@flawedmatrix What are the cluster objects beings used for?
From this https://github.com/telekom/cluster-api-ipam-provider-in-cluster/blob/1144752b5b72118d13c4e9f9539af6c468ca2def/internal/controllers/ipaddressclaim.go#L70-L89, seems like we are adding a watch on the Cluster object, and any update on the Cluster object is requeuing all the IPAddressClaims for the Cluster (irrespective of any relevant changes on the pool/claims themselves)

My suggestion is to remove the watch altogether and continue using the other watches to reconcile the claims. @schrej wdyt?

schrej:
We've added a feature that pauses claim reconciliation when a cluster is paused. The watch is there to reconcile again once the cluster is unpaused. If we want to keep the feature, we need to keep the watch as well.

I somehow missed this PR and opened another one just now :D

PR #123, opened 2023-04-18
Remove subnet, first and last from the pool spec and create v1alpha2 Pool

### LINKED ISSUE #71: Pools fail to render the first/last - start/end IP addresses in the kubernetes cli

christianang:
Following what was discussed in #71. This PR removes subnet, first, and last from the pool spec in favor of a user using the new addresses field.

This PR will support a new v1alpha2 version for the InClusterIPPool and GlobalInClusterIPPool. In addition it will convert v1alpha1 resources to v1alpha2 resource using a conversion webhook. First and Last will be converted to use an addresses range in the new pool spec.

The controller and default/validation webhooks are configured to only support v1alpha2 pools and it is expected that the conversion webhook will convert v1alpha1 resources to v1alpha2 beforehand.

We have also added conversion-gen to the Makefile to support future conversions.

Fixes #71


christianang:
@schrej I have rebased this PR and included the new changes from the pool PR. This should be ready for review. We will probably serialize PRs a bit and leave the other PRs alone until this one is merged to avoid multiple rebases since this PR is fairly significant.

api/v1alpha1/inclusterippool_conversion.go
```diff
@@ -0,0 +1,109 @@
...
+	if err := Convert_v1alpha2_InClusterIPPool_To_v1alpha1_InClusterIPPool(src, dst, nil); err != nil {
+		return err
+	}
+
+	restored := &InClusterIPPool{}
+	if ok, err := utilconversion.UnmarshalData(src, restored); err != nil || !ok {
+		return err
+	}
+
+	dst.Spec = restored.Spec
```
schrej, line 53: I don't think this will work. It completely ignores any modifications made to the resource using v1alpha2, doesn't it?

IF we want to do proper api versioning, and want to implement conversion, I think it needs to be a bit more sophisticated. It should be possible to properly convert between the versions, since no information is removed. The layout just changes, becomes narrower.
`.spec.first` and `.spec.last` can only be used when a single range of addresses is specified. Same goes for subnet. So we should only fill those values when converting alpha2->alpha1 when `.spec.addresses` contains only one entry. Otherwise we'll just use `.spec.addresses`.
The only issue is: we need to know which format was used with alpha1. We can simply store that in an annotation (only if it's first/last or subnet - addresses is the default) when converting 1->2 and then _try_ to respect that annotation when converting 2->1.

Generally adding annotations to the hub version isn't very pretty, since those will remain even if conversion is removed at some point, but there probably is no better way to handle it here.

schrej (changes requested):

api/v1alpha1/inclusterippool_conversion.go
adobley: I suggest we just go with using `.addresses` for 2->1. It feels like a decent effort to get the conversion back to `.first` and `.last` correct and I'm not sure what it buys us other than some presentation to a user. The behavior is the same regardless of which field is used to represent the data. Any concerns that you have with us going ahead with that approach?

schrej: I was thinking about automated tooling that's still consuming alpha1. Tools like flux would fetch the resource, compare it to their known state and compare. If it deviates, it would apply again. Or tooling that applies and then checks whether it's correct. If we just use `.addresses`, what you write wouldn't be what you get back from the api server.

On the other hand, it might not be worth the annotation that we'll have to clean up at some point in the future (not too big of a deal though).

christianang: Thinking about it more deeply with Aidan and team. If we want to support returning the values they had in v1 in the same format they inputted the data in v1 we should only do that if they last updated the data in v1 i.e if they initially apply in v1, and make updates in v1, we should return what they had in v1. If they apply in v1, but make an update in v2, we should start returning it as v2 instead of figuring out what they initially did in v1 and transforming v2 data.

Thinking about what this might be implemented as. Ideally we could remove the annotation if they make an update in v2, but not sure if there is a way to do that cleanly. I was thinking it could be removed in the webhook, but v1 is already converted to v2 at that point.

The other implementation would be to conditionally apply the `restored.Spec` if the up-converted restored spec matches the current spec (this means the annotation is still valid and therefore we can return it).

schrej: We could just remove the annotation once we remove v1alpha1 entirely. It doesn't really matter whether it's there or not.

I didn't get the last paragraph. How would you check that?

christianang: The thought was we could take what was in the annotation and convert it to v1alpha2 to compare it apples to apples with the current spec. If it matches we use what was in the annotation and if it doesn't we return what is in the current spec. To me this is okay because if they were to make an update using v1alpha2 I'm assuming then they wouldn't want to downconvert to use the first and last format.

And because this is somewhat confusing to talk about here is psuedo code on what I'm thinking:

```go
restoredV2 := &v1alpha2.InClusterIPPool{}
if err := Convert_v1alpha1_InClusterIPPool_To_v1alpha2_InClusterIPPool(restored, restoredV2, nil); err != nil {
		return err
}

if restoredV2.spec.addresses == src.spec.addresses &&
		restoredV2.spec.prefix == src.spec.prefix &&
		restoredV2.spec.gateway == src.spec.gateway {
		dst.Spec = restored.Spec
}
```

schrej: Ah, now it makes sense to me. Yes, we can also go with that approach. I'll let you decide.

christianang: Okay, I've updated the v2alpha1 to v1alpha1 convert to do the above approach. Thanks.

api/v1alpha1/conversion.go
```diff
@@ -0,0 +1,59 @@
...
+
+	if in.Subnet != "" {
+		prefix, err := netip.ParsePrefix(in.Subnet)
+		if err != nil {
+			return field.Invalid(field.NewPath("spec", "subnet"), in.Subnet, err.Error())
+		}
+
+		prefixRange := netipx.RangeOfPrefix(prefix)
+		if in.First == "" {
+			in.First = prefixRange.From().Next().String() // omits the first address, the assumed gateway
```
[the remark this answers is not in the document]
schrej: While looking at this I was thinking whether we're handling CIDRs in the `addresses` field correctly, and what "correctly" means here.

In IPv4 you'll typically have a network address (the first one) and a broadcast address (the last one) within a full subnet (which can be specified using a CIDR string) ([RFC](https://datatracker.ietf.org/doc/html/rfc1812#section-4.2.3.1) that confirms this). This is not always the case, in our internal setup there are no network addresses needed for example. 
For IPv6, the first address seems to be reserved as a special anycast address (anycast is also replacing broadcast addresses) as per [RFC](https://www.rfc-editor.org/rfc/rfc4291#section-2.6.1).

So maybe we should exclude the first and last addresses for IPv4, and the first for IPv6?

There is even additional rules for very small networks (/31 for v4, /127 for v6) that get rid of those reserved addresses.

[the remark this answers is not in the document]
tylerschultz: I'm thinking about the case that looks like this:
```
Gateway: 192.168.0.1
Prefix: 16
Addresses:
- 192.168.0.0/24
- 192.168.200.0/24
```
I agree, given what you've said, it makes sense that we remove 192.168.0.0 and 192.168.255.255.
It seems we should not remove 192.168.200.0 nor 192.168.200.255.

The key part being that CIDRs in the addresses are ranges, not network definitions.

I propose we should remove the derived subnet's network address for both IPv4 and iPv6, and broadcast address for IPv4.

This seems like incorrect behavior in v1 and v2, and perhaps should be addressed (pun intended) in a separate issue/PR? WDYT?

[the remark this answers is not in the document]
christianang: @schrej any thought on what @tylerschultz commented above. Also happy to implement it in this PR if it feels like that makes sense.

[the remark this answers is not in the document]
christianang: Sounds good. I made the following issue https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/issues/165, and we can follow up with a PR to fix this issue after this PR is merged. I will resolve this thread and I think at this point we are happy with this PR, unless you have any other feedback.

schrej:
/lgtm
/approve

Thank you! And sorry that this was a bit slow, lost track of it at some point.

PR #124, opened 2023-04-19
protect pools with IP allocations from deletion

### LINKED ISSUE #114: Prevent pool deletion if IPAddress exists for pool
# Problem

I noticed that a pool can be deleted even if IPAddresses have been allocated for that pool. This seems undesirable.

# Possible solutions

Option 1: Add a finalizer to the pool that can be used to ensure the pool isn't removed until all IPAddresses are deleted. Drawback here is that the pool is marked for deletion and this may be an undesirable state if a user decides that they can't clean up the IPAddresses and want to continue using the pool.

Option 2: Guard deletion of the pool with the webhook. More complex then a finalizer, but prevents them from potentially entering a bad state.

I'm leaning towards implementing option 2 if we do this, but open to hearing opinions here.

adobley:
Adds a finalizer to prevent pool deletion when IPs are still allocated

Adds a delete webhook check that will ensure pools with allocated IPs cannot be removed

This change relies on work in #116, when that is merged we can rebase this.

Fixes #114

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -77,6 +77,7 @@ webhooks:
     operations:
     - CREATE
     - UPDATE
+    - DELETE
     resources:
     - inclusterippools
   sideEffects: None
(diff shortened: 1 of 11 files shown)


schrej:
#116 is in now!

PR #125, opened 2023-04-19
Webhook Validate updated pool does not remove in use IPs

### LINKED ISSUE #117: Addressing downscaling pools with addresses in range
# Context

We see that pools can be updated without regard for what IPs are in use. This can lead to situations where IPs can become out of range of the pool's configuration.

# Potential Solution

We are thinking of adding validation in the webhook that would check if there is an IP address that is already allocated before allowing it to be removed from the pool. This continues to allow configuration of the pool, but prevents an IP Address that is in use to be removed from the pool.

We are also thinking of adding an `outOfRange` status count on the pool status (similar to #112) to expose potential issues if it already happens to have out of range IPs. This is mostly considering the case of updates to the pool that may have occurred without the webhook validation.

In addition to the `outOfRange` status count on the pool we could also add an outOfRange status condition on the IPAddress.

adobley:
In the case that the a pool is still updated and has out of range IPs display the count in the pool status

Relies on #116, we can update this when that is merged

Fixes #117, with the exception of `In addition to the outOfRange status count on the pool we could also add an outOfRange status condition on the IPAddress.`

This last bit would require changes to the IPAddress resource in core CAPI. The Status field is missing on the IPAddress objects. We think this is okay for now and we can open a new issue for that if we still think it is valuable.

### DIFF
--- a/api/v1alpha1/inclusterippool_types.go
+++ b/api/v1alpha1/inclusterippool_types.go
@@ -59,6 +59,11 @@ type InClusterIPPoolStatusIPAddresses struct {
 	// Used is the count of allocated IPs in the pool.
 	// Counts greater than int can contain will report as math.MaxInt.
 	Used int `json:"used"`
+
+	// Out of Range is the count of allocated IPs in the pool that is not
+	// contained within spec.Addresses.
+	// Counts greater than int can contain will report as math.MaxInt.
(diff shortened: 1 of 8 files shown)


schrej (approved):

schrej:
For some reason Actions aren't running for this one 🤔

Issue #140, opened 2023-05-09
clusterctl move fails because of delete webhook


christianang:
# Context

When clusterctl move runs it attempts to delete everything off the source cluster. This currently fails because it may attempt to delete the pool before it deletes all the ipaddresses and there doesn't seem to be a way to control delete order. So it will error on the validate delete webhook, which #124 added functionality to check that there are no in use IPs before allowing a pool to be deleted.

# Potential solution

For future cluster-api releases (1.5+), there was an annotation added that a validate delete webhook can look for to potentially allow the validation to be skipped called: `clusterctl.cluster.x-k8s.io/delete-for-move` (via https://github.com/kubernetes-sigs/cluster-api/pull/8322).

For the current cluster-api release and below (1.4.x), there doesn't seem to be a good way to do this so I think we need to add our own annotation that allows the delete validation to be skipped, which a user could have on their pool before a move e.g `ipam.cluster.x-k8s.io/skip-validate-delete-webhook`.

PR #141, opened 2023-05-09
:bug: Add skip validate delete webhook annotation

christianang:

This PR adds a `ipam.cluster.x-k8s.io/skip-validate-delete-webhook` annotation that can be added to pools to skip delete validation. This is necessary for clusterctl move to work because it attempts to delete everything off the source cluster and it cannot control the order of deletion so it may delete pools before ipaddressclaims.

Related to #140

### DIFF
--- a/internal/webhooks/inclusterippool.go
+++ b/internal/webhooks/inclusterippool.go
@@ -20,6 +20,13 @@ import (
 	"github.com/telekom/cluster-api-ipam-provider-in-cluster/pkg/types"
 )
 
+const (
+	// SkipValidateDeleteWebhookAnnotation is an annotation that can be applied
+	// to the InClusterIPPool or GlobalInClusterIPPool to skip delete
+	// validation. Necessary for clusterctl move to work as expected.
+	SkipValidateDeleteWebhookAnnotation = "ipam.cluster.x-k8s.io/skip-validate-delete-webhook"
+)
+
 func (webhook *InClusterIPPool) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	err := ctrl.NewWebhookManagedBy(mgr).
 		For(&v1alpha1.InClusterIPPool{}).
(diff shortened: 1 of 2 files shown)


internal/webhooks/inclusterippool.go
```diff
@@ -20,6 +20,13 @@ import (
 	"github.com/telekom/cluster-api-ipam-provider-in-cluster/pkg/types"
 )
 
+const (
+	// SkipValidateDeleteWebhookAnnotation is an annotation that can be applied
+	// to the InClusterIPPool or GlobalInClusterIPPool to skip delete
+	// validation. Necessary for clusterctl move to work as expected.
+	SkipValidateDeleteWebhookAnnotation = "ipam.cluster.x-k8s.io/skip-validate-delete-webhook"
```
schrej, line 27: Can we import the package and use the constant that's defined here instead? https://github.com/pkbhowmick/cluster-api/blob/main/cmd/clusterctl/api/v1alpha3/annotations.go

schrej (changes requested):
lgtm otherwise

internal/webhooks/inclusterippool.go
christianang: The delete-for-move annotation isn't in a released version of cluster-api yet, so this won't work for users using cluster-api v1.4.x. Unless the thinking is that a user would add the delete-for-move annotation manually, as was my original intention with this `skip-validate-delete-webhook` annotation.

schrej: I just didn't want to redefine the constant. If it's not release on CAPI yet we can either wait, or if we need to get this in sooner we can keep the duplicated one, and maybe replace it later.

christianang: Personally, I would prefer if we added this and replaced it later because at the moment clusterctl move doesn't work. However, I'm happy to go with whatever final decision you have if you have a strong opinion. I understand the desire to not rework this between versions too.

schrej: Let's replace the constant once it's released (if we remember to 😄 )

PR #144, opened 2023-05-19
🐛 Skip claim reconcile when cluster unretrievable

### LINKED ISSUE #143: Claims with cluster label should not be reconciled when cluster cannot be retrieved
When using Velero to do a backup/restore, we see that claims are getting reconciled when the paused cluster has yet to be restored. We see that the cluster is always restored last, due to the dependency tree.

Our use case is Velero specific, but we think this is a more general bug too. If a claim is linked to a cluster and it can't be retrieved, we think it should skip reconciliation until the cluster can be evaluated.

tylerschultz:

If a claim has a cluster name label, and the cluster cannot be retrieved, do not reconcile the claim. Wait for a reconcile loop when the cluster can be evaluated for Paused state.

Fixes #143


internal/controllers/ipaddressclaim.go
```diff
@@ -160,17 +160,27 @@ func (r *IPAddressClaimReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		cluster, err := clusterutil.GetClusterFromMetadata(ctx, r.Client, claim.ObjectMeta)
+		if err != nil && !apierrors.IsNotFound(err) {
+			log.Error(err, "error fetching cluster linked to IPAddressClaim")
+			return ctrl.Result{}, err
+		}
+
+		if err != nil && apierrors.IsNotFound(err) {
+			log.Info("IPAddressClaim linked to a cluster that is not found, unable to determine cluster's paused state, skipping reconciliation")
+			return ctrl.Result{}, nil
+		}
```
schrej, line 173: What do you think about this instead?

```suggestion
		if err != nil {
		  if apierrors.IsNotFound(err) {
			  log.Info("IPAddressClaim linked to a cluster that is not found, unable to determine cluster's paused state, skipping reconciliation")
			  return ctrl.Result{}, nil
		  }
			log.Error(err, "error fetching cluster linked to IPAddressClaim")
			return ctrl.Result{}, err
		}
```

tylerschultz: Sure, I like it like that too. Resolved.

schrej (approved):

PR #159, opened 2023-06-12
:bug: Migrate modules to right path

rikatz:
Today the code was migrated keeping the old reference to the original repo. This becomes a problem while trying to use/import some helper functions, etc :)


schrej:
Could you update https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/blob/main/.golangci.yml#L58 so the linter is happy?

PR #160, opened 2023-06-13
:seedling: Add tests for non existant pools

rikatz:
There is a missing test on reconciliation for when the pool does not exist. This could lead to panic due to pool being nil and still executing the `annotations.HasPaused(pool)`

OTOH, with this change we don't execute r.reconcileDelete anymore if the pool does not exists, which may be an error (need to be checked)


internal/controllers/ipaddressclaim.go
```diff
@@ -226,6 +226,12 @@ func (r *IPAddressClaimReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		pool = gicippool
 	}
 
+	if pool == nil {
+		err := errors.New("pool not found")
+		log.Error(err, "the referenced pool could not be found")
+		return ctrl.Result{}, nil
+	}
+
 	if annotations.HasPaused(pool) {
```
schrej, line 235: As you mentioned, this prevents executing `reconcileDelete` if the pool is not found. While the current implementation would prevent getting into a state where the Finalizer is set on the claim without the pool existing, older Claims could end up in that state. In addition, manually deleting Finalizers could also lead to such a situation.

How about adding a `pool != nil` check here, and moving the `pool == nil` below the call to `reconcileDelete`?

schrej (changes requested):

internal/controllers/ipaddressclaim.go
rikatz: sounds good, just changed here :)

schrej:
Thanks!

/lgtm
/approve

PR #167, opened 2023-06-29
🐛 Update cluster paused detection

### LINKED ISSUE #136: IPAddressClaim controller should detect cluster paused consistently
Changes were recently made to the IPAddressClaim controller to prevent reconciliation when the `Cluster` object is paused. The `Cluster` object has a `spec.Paused` property, and can also have a Paused annotation. The code copied from CAPV and placed into the IPAddressClaim controller looks at the `Cluster`'s `spec.Paused` property during updates, and looks at the `Cluster`'s paused annotations during creates. It seems the Update and Create functions should both look at both the annotation and the property.

https://github.com/telekom/cluster-api-ipam-provider-in-cluster/blob/main/internal/controllers/ipaddressclaim.go#L76-L88

We thought this was odd and @srm09 agreed. We opened this issue on CAPV as a result of our discussion. 

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1890

Sagar suggests this paused detection predicate should be a shared library function owned by either CAPV or CAPI, and all controllers should detect paused in the same. This issue is a reminder to update the IPAddressClaim controller when the CAPV issue is resolved.

flawedmatrix:
Clusters can be paused by annotation or by spec.paused. If either are true, skip claim reconcilliation that are associated to those claims.

Fixes #136

### DIFF
--- a/internal/controllers/ipaddressclaim.go
+++ b/internal/controllers/ipaddressclaim.go
@@ -93,10 +93,11 @@ func (r *IPAddressClaimReconciler) SetupWithManager(ctx context.Context, mgr ctr
 				UpdateFunc: func(e event.UpdateEvent) bool {
 					oldCluster := e.ObjectOld.(*clusterv1.Cluster)
 					newCluster := e.ObjectNew.(*clusterv1.Cluster)
-					return oldCluster.Spec.Paused && !newCluster.Spec.Paused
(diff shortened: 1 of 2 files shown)


tylerschultz:
I worked on this commit, so I should recuse myself. But, FWIW, LGTM!

PR #171, opened 2023-07-07
🐛 Controller does not allocate reserved addresses

### LINKED ISSUE #165: Controller shouldn't allocate IP addresses that are "reserved" for the subnet
# Context

Broken out of the conversation started in https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/pull/123#discussion_r1200845468.

There are ip addresses that shouldn't be given out that are reserved for special cases in the subnet. This is typically the network address, the broadcast address, and the gateway address.

# Proposed Solution

If I defined the following pool:
```
Gateway: 192.168.0.1
Prefix: 16
Addresses:
- 192.168.0.0/16
```

The controller should not allocate: 192.168.0.0 (network address), 192.168.0.1 (gateway address), 192.168.255.255 (broadcast address). It would therefore give out 192.168.0.2-192.168.255.254.

If a user wants to allocate reserved addresses i.e ignore the above functionality except for gateway, a user can set a new flag called `allocatedReservedAddresses`, which would allocate 192.168.0.0-192.168.255.255 (except for 192.168.0.1). If the user doesn't want to reserve the gateway they should not set the gateway.

adobley:
The validation code continues to accept pools that contain the network (the first address in the inferred subnet) and broadcast (the last address in the inferred subnet) addresses, but the controller will not allocate these addresses. When the pool is configured with IPv6, the anycast address (the first address in the inferred subnet) will not be allocated.

This PR also refactors the claim reconciler. Previously, the controller would continue to reconcile (and ultimately no-op) despite the pool not being found. The new spec flag,`AllocateReservedIPAddresses` exposed this oddness, prompting some refactoring.

Fixes #165


internal/controllers/ipaddressclaim.go
```diff
@@ -366,17 +378,11 @@ func (r *IPAddressClaimReconciler) clusterToIPClaims(a client.Object) []reconcil
...
+func buildAddressList(addressesInUse []ipamv1.IPAddress) []string {
+	addrStrings := make([]string, len(addressesInUse))
 	for i, address := range addressesInUse {
 		addrStrings[i] = address.Spec.Address
 	}
-
-	if gateway != "" {
-		addrStrings = append(addrStrings, gateway)
-	}
-
```
tylerschultz, line 379: We no longer consider the gateway in use, but with this change set the gateway is not in available list.

internal/controllers/inclusterippool_test.go
```diff
@@ -94,34 +94,47 @@ var _ = Describe("IP Pool Reconciler", func() {
...
 			Entry("When there are 2 claims with gateway in range - InClusterIPPool",
-				"InClusterIPPool", []string{"10.0.0.10-10.0.0.20"}, "10.0.0.10", 10, 2, 8),
+				"InClusterIPPool", 24, []string{"10.0.0.10-10.0.0.20"}, "10.0.0.10", 10, 2, 8),
 			Entry("When there is 1 claim with gateway outside of range - InClusterIPPool",
-				"InClusterIPPool", []string{"10.0.0.10-10.0.0.20"}, "10.0.0.1", 11, 1, 10),
+				"InClusterIPPool", 24, []string{"10.0.0.10-10.0.0.20"}, "10.0.0.1", 11, 1, 10),
+			Entry("When the addresses range includes network addr, it is not available - InClusterIPPool",
+				"InClusterIPPool", 24, []string{"10.0.0.0-10.0.0.1"}, "10.0.0.2", 1, 1, 0),
+			Entry("When the addresses range includes broadcast, it is not available - InClusterIPPool",
+				"InClusterIPPool", 24, []string{"10.0.0.254-10.0.0.255"}, "10.0.0.1", 1, 1, 0),
```
tylerschultz, line 109: When a reserved address is in the `Spec.Addresses`, it is not considered used, nor free.

tylerschultz:
Once again, I worked on this commit, so I should recuse. FWIW, LGTM. @adobley and I re-reviewed it after some time away and we think it's good to go.

adobley:
Fixed a misspelling, should be good to go.

schrej:
lgtm, but the linter is complaining about one spelling mistake.
Can you check why CLA is unhappy?

Sorry that it took so long, a lot going on.

adobley:
Fixed the typo and the CLA fixed itself. Strange.

internal/controllers/ipaddressclaim.go
```diff
@@ -211,57 +211,60 @@ func (r *IPAddressClaimReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		}
+		return icippool, nil
+	} else if claim.Spec.PoolRef.Kind == globalInClusterIPPoolKind {
+		gicippool := &v1alpha2.GlobalInClusterIPPool{}
+		if err := r.Client.Get(ctx, types.NamespacedName{Name: claim.Spec.PoolRef.Name}, gicippool); err != nil {
+			return nil, err
+		}
+		return gicippool, nil
 	}
+	return nil, fmt.Errorf("Unknown pool type: %s", claim.Spec.PoolRef.Kind)
```
schrej, line 257: Error strings should be lowercase (the linter complains).

adobley: Fixed! Thanks!

PR #172, opened 2023-07-10
📖 Update README.md

### LINKED ISSUE #169: The provider configuration to add to clusterctl.yaml is wrong
The provider URL is no longer on telekom but has moved to kubernetes-sigs, so should be:

```
providers:
  - name: in-cluster
    url: https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/releases/latest/ipam-components.yaml
    type: IPAMProvider
```

otherwise clusterctl init errors with:

```
Fetching providers
Error: failed to get provider components for the "in-cluster" provider: target namespace can't be defaulted. Please specify a target namespace
```

agtogna:

The provider URL to be used in `clusterctl init` has changed since the move from telekom to kubernetes-sigs

**Which issue(s) this PR fixes**

Fixes #169

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,17 @@
 
 This is a IPAM provider for Cluster API that manages pools of IP addresses using Kubernetes resources. It serves as a reference implementation for IPAM providers, but can also be used as a simple replacement for DHCP.
 
+## Cluster API setup via clusterctl
+
+This release comes with clusterctl support. Since it's not added to the list of built-in providers yet, you'll need to add the following to your `~/.cluster-api/clusterctl.yaml` if you want to install it using `clusterctl init --ipam incluster`:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,17 @@
 
 This is a IPAM provider for Cluster API that manages pools of IP addresses using Kubernetes resources. It serves as a reference implementation for IPAM providers, but can also be used as a simple replacement for DHCP.
 
+## Cluster API setup via clusterctl
```
schrej, line 5: Can we remove "Cluster API" here? A bit misleading since we only want to setup the provider instead of the entirety Cluster API 😄

schrej (commented):
LGTM apart from one nit

PR #176, opened 2023-07-26
:bug: Make sure the ListAddressesInUse doesn't take the namespace into account for GlobalInClusterIPPools

### LINKED ISSUE #175: :bug: GlobalInClusterIPPool does not guarantee the unique IP across multiple k8s namespaces
When using the `GlobalInClusterIPPool` (the CRD is not namespace scoped) and referencing it by creating various `IpAddressClaim` in various namespaces, the generated `IpAdress` crs then contain IP are not unique. They basically start the sequence for each namespace again.

jkremser:
When calling `ListAddressesInUse` the namespace shouldn't be set to the same value as `IpAddressClaim`'s ns in case of using the `GlobalInClusterIPPool`. Otherwise, `IpAddressClaim`s created in different namespaces, pointing to the same `GlobalInClusterIPPool` (which itself isn't namespaced) will yield the duplicate IP addresses.

**Which issue(s) this PR fixes** 
Fixes #175


jkremser:
I tried the `main` and it works as expected, the issue is present in the latest released image (`:v0.1.0-alpha.2`) so I am closing this issue and pr. May I ask if you plan to cut a new release? Depending on that `:main` container image can be dangerous for us.

PR #182, opened 2023-08-22
✨ Allow address exclusion

### LINKED ISSUE #180: Handle reserved Ips withing an IP range
Wen are trying to integrate this provider as part of our infrastructure in Proxmox.

But we couldn't find any way of handling reserved IPs.

As we use this IPAM provider to get ip addresses to set to our machines, but there's no way of setting an already 
existent IP. 
which means, that would end up in a conflicts.

for example we have an IPrange: 10.10.10.1/24 and gateway: 10.10.10.1

In our range, we would have some Machines running , e.g. jumphosts, CI, dhcp server, or VMs used for maintenance.

so, if we have a manual machine provisioned to this network with ip e.g. `10.10.10.3`,

How to tell IPAM provider to no use this IP?

Noroth:
Currently, the only way to exclude certain IP addresses is to specify multiple ranges inside the `addresses` field in the `InClusterIPPoolSpec`.

To enhance this, a new optional field will be added, where another list of IP addresses can be provided. These will be excluded during the allocation of new IP addresses.

Fixes #180


internal/webhooks/inclusterippool.go
```diff
@@ -212,6 +212,21 @@ func (webhook *InClusterIPPool) validate(_, newPool types.GenericInClusterPool)
...
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "excludedAddresses"), address, "provided address is not a valid IP, range, nor CIDR"))
+			continue
+		}
+		from := ipSet.Ranges()[0].From()
+		hasIPv4Addr = hasIPv4Addr || from.Is4()
+		hasIPv6Addr = hasIPv6Addr || from.Is6()
+	}
+
+	if hasIPv4Addr && hasIPv6Addr {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "excludedAddresses"), newPool.PoolSpec().ExcludedAddresses, "provided addresses are of mixed IP families"))
```
schrej, line 227: We need to make sure that the family of `excludedAddresses` matches the family of `addresses`, and not that the family is the same for all. This approach _should_ work, but it would be good to be a bit more explicit about it.

schrej (changes requested):
One minor thing, looks good apart from that. Tests are failing and the linter is unhappy, could you take a look?

internal/webhooks/inclusterippool.go
Noroth: No problem, will add an additional check. 

I initially thought that it wouldn't really matter as we are creating a set of IPs from the provided addresses and then just removing whatever is being found in the excluded addresses. Therefore if you attempt to remove ipv6 addresses from ipv4 addresses and vice versa, nothing will happen.

Noroth:
Seems like a test timed out. I tried to reproduce it locally but tests are never failing for me

Noroth:
/cc @schrej 

I'm ready, could you please take another look?

schrej (approved):
/lgtm

schrej:
And thank you, by the way! 🎉

Noroth:
No problem, happy to contribute. Was always looking for a task that I could work on in the k8s environment 😁

PR #203, opened 2023-11-09
:warning: Generic claim reconciler

p-strusiewiczsurmacki-mobica:

This PR introduces generic claim reconcile - an interface that makes underlying IP address pools transparent for the controller and therefore can be used for easy integration of other IPAM providers.

Changes in this PR were created with @schrej


internal/controllers/ipaddressclaim.go
```diff
@@ -21,23 +21,16 @@ import (
 	"fmt"
 
 	"github.com/pkg/errors"
+	"golang.org/x/exp/slices"
```
rikatz, line 24: If we bump to go v1.21 we can use the builtin slices: https://pkg.go.dev/slices#ContainsFunc

pkg/ipamutil/reconciler.go
```diff
@@ -0,0 +1,281 @@
...
+	SetupWithManager(context.Context, *ctrl.Builder) error
+	// ClaimHandlerFor is called during reconciliation to get a ClaimHandler for the reconciled [ipamv1.IPAddressClaim].
+	ClaimHandlerFor(client.Client, *ipamv1.IPAddressClaim) ClaimHandler
+}
+
+// ClaimHandler knows how to allocate and release IP addresses for a specific provider.
+type ClaimHandler interface {
+	// FetchPool is called to fetch the pool referenced by the claim. The pool needs to be stored by the handler.
+	FetchPool(ctx context.Context) (*ctrl.Result, error)
+	// EnsureAddress needs to make sure an ip address is allocated for the claim and the spec of the provided [ipamv1.IPAddress] is correct.
```
rikatz, line 71: This comment is a bit confusing. As a maintainer of another pool, what should I do? Maybe:

"EnsureAddress is called to verify if the spec of an address matches the claim instance".
Also, is the idea of Ensure to allocate an IPAddress if nothing is allocated? (I probably should keep reading the PR, just trying to make sense of things here :) )

pkg/ipamutil/reconciler.go
```diff
@@ -0,0 +1,281 @@
...
+					return annotations.IsPaused(oldCluster, oldCluster) && !annotations.IsPaused(newCluster, newCluster)
+				},
+				CreateFunc: func(e event.CreateEvent) bool {
+					cluster := e.Object.(*clusterv1.Cluster)
+					return !annotations.IsPaused(cluster, cluster)
+				},
+			}),
+		)
+
+	if err := r.Adapter.SetupWithManager(ctx, b); err != nil {
```
rikatz, line 103: If SetupWithManager is not set, this will panic. Same for adapter.
Probably it is worth to check if both are null before calling them

rikatz: Also, if Adapter is required, this SetupWithManager should return an error if Adapter is null

pkg/ipamutil/reconciler.go
```diff
@@ -0,0 +1,281 @@
...
+	// Fetch the IPAddressClaim
+	claim := &ipamv1.IPAddressClaim{}
+	if err := r.Get(ctx, req.NamespacedName, claim); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	// Check if the owning cluster is paused
```
rikatz, line 122: Isn't this already verified on predicate? (maybe not for deleteFunc, but for others I can see it is verified).

pkg/ipamutil/reconciler.go
```diff
@@ -0,0 +1,281 @@
...
+			}
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, errors.Wrap(err, "failed to fetch pool")
+	}
+
+	if pool := handler.GetPool(); pool != nil && annotations.HasPaused(pool) {
+		log.Info("IPAddressClaim references Pool which is paused, skipping reconciliation.", "IPAddressClaim", claim.GetName(), "Pool", pool.GetName())
+		return ctrl.Result{}, nil
+	}
```
rikatz, line 172: Is there a reason for FetchPool and IPPool being separate operations? IIUC on both you get the pool, then you should return the pool value.

rikatz: I mean, the FetchPool() could be called once, return an IPPool and then you use it. Unless the concern is like, people calling FetchPool on a crazy way and always forcing its reconciliation.

pkg/ipamutil/reconciler.go
```diff
@@ -0,0 +1,281 @@
...
+
+	// If the claim is marked for deletion, release the address.
+	if !claim.ObjectMeta.DeletionTimestamp.IsZero() {
+		if res, err := handler.ReleaseAddress(); err != nil || res != nil {
```
rikatz, line 176: this `res != nil` check, I'm wondering what happens if my ReleaseAddress returns an empty instance of Result, like

```
return ctrl.Result{}, nil
```

This means that the instance exists but it is empty (not null) IIRC. I may be wrong, tho! lmk if this makes sense, if we are not returning a null instance (but an empty) probably we will get stuck into this problem? And never reach reconcileDelete, or reach just after another run

internal/controllers/ipaddressclaim.go
```diff
@@ -59,79 +52,72 @@ const (
 	ProtectAddressFinalizer = "ipam.cluster.x-k8s.io/ProtectAddress"
 )
 
-// IPAddressClaimReconciler reconciles a InClusterIPPool object.
-type IPAddressClaimReconciler struct {
-	client.Client
-	Scheme *runtime.Scheme
+type genericInClusterPool interface {
```
rikatz, line 55: Real nit for this file name: should it be renamed to incluster.go instead of ipaddressclaim? I first got confused on what this means, then I got that here now you have your IPAddress Claim implementation for the InCluster controller

rikatz:
sorry for the amount of comments and the delay!!!
I've left mostly some concerns and some nits, overall this is great!! Happy to see this moving on!!

pkg/ipamutil/reconciler.go
```diff
@@ -0,0 +1,281 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	ctrlhandler "sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
```
rikatz, line 28: the finalizer usage here is a bit confusing to me.

When is ReleaseAddress used? At which time? And Protect?

Usually it helps me to describe on a more verbose way at some comment something like 

"When a new claim arrives, it will do X, then Y, then add a finalizer to guarantee this or that"

But this can be a followup!!!! No need to sweat on it

pkg/ipamutil/reconciler.go
schrej: It's supposed to ensure an address is allocated, and that it's set correctly in the IPAddress.spec (address, gateway, prefix). I agree the comment is not clear enough.

pkg/ipamutil/address.go
```diff
@@ -28,8 +28,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-// NewIPAddress creates a new ipamv1.IPAddress with references to a pool and claim.
-func NewIPAddress(claim *ipamv1.IPAddressClaim, pool client.Object) ipamv1.IPAddress {
+// newIPAddress creates a new ipamv1.IPAddress with references to a pool and claim.
```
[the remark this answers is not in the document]
schrej: When using the generic reconciler, this function doesn't need to be called anymore, which is probably why it was made private. I think we can revert that though.

pkg/ipamutil/reconciler.go
schrej: iirc this just evolved this way. The idea was that the handler stores the pool on it's own and directly accesses it when it's needed. That way we don't need generics or type conversions. But since the reconciler needs to access the pools metadata in some cases I added that `GetPool()` method as well.
I think we can change FetchPool() to return the pool as well, and remove this.

pkg/ipamutil/reconciler.go
```diff
@@ -0,0 +1,281 @@
...
+	address := &ipamv1.IPAddress{}
+	namespacedName := types.NamespacedName{
+		Namespace: claim.Namespace,
+		Name:      claim.Name,
+	}
+	if err := r.Client.Get(ctx, namespacedName, address); err != nil && !apierrors.IsNotFound(err) {
+		return ctrl.Result{}, errors.Wrap(err, "failed to fetch address")
+	}
+
+	if address.Name != "" {
```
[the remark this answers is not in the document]
schrej: No, `address` is instantiated with an empty `IPAddress` above and will only be filled by the `Get()` call later.

internal/controllers/ipaddressclaim.go
schrej: Valid point. I kept the file name since it's the custom logic that reconciles claims, even though it's using a generic implementation for a lot of it. When you are familiar with kubebuilder projects, this is where you would expect to find that logic imo.

pkg/ipamutil/reconciler.go
schrej: I think there is a lot of disagreement around Finalizer Naming. To me a Finalizer takes care of a specific action/function, and is removed after that task is done. When the same controller handles multiple Finalizers it's technically not necessary to have more than one. I think that's the reason while there is often just some generic finalizer to prevent deletion which gets removed after everything is done.
In this case the Finalizers serve different purposes though. The `ProtectAddress` one prevents deletion until the Claim is deleted. The `ReleaseAddress` one cleans up the allocated address after the claim was deleted. With the in-cluster provider this of course happens immediately, with e.g. Infoblox this can be asynchronous.
Whether this approach makes sense is debatable, especially since the `ProtectAddress` finalizer should also prevent the `ReleaseAddress` finalizer from executing, as we don't want to deallocate the address while it is still in use.

schrej:
> sorry for the amount of comments

Please don't apologise for a thorough review, thanks for making time for it!

internal/controllers/ipaddressclaim.go
p-strusiewiczsurmacki-mobica: Updated go to 1.21

pkg/ipamutil/address.go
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: I've made `NewIPAddress` exported once again.

pkg/ipamutil/reconciler.go
p-strusiewiczsurmacki-mobica: Changed comment to: EnsureAddress is called to make sure that the IPAddress.Spec is correct and the address is allocated.

pkg/ipamutil/reconciler.go
p-strusiewiczsurmacki-mobica: Added check for adapter == nil.
Adapter is an interface, so I think that anything passed has to define `SetupWithManager`, as it won't fulfill interface otherwise. Or is there something I am missing here?

pkg/ipamutil/reconciler.go
p-strusiewiczsurmacki-mobica: Changed FetchPool to return pool, res and err. Deleted GetPool.

pkg/ipamutil/reconciler.go
p-strusiewiczsurmacki-mobica: You're right. Changed that to just check error.

pkg/ipamutil/reconciler.go
schrej: Ah, I remember this now: the predicate regarding cluster primarily exists to notice when the cluster is unpaused again. It's not sufficient to filter on that level to avoid reconciliations when the cluster is paused.

pkg/ipamutil/reconciler.go
rikatz: ok, makes sense. I was looking into the perspective of incluster and keep forgetting that at some points the deletion may be an async operation

internal/controllers/ipaddressclaim.go
```diff
@@ -19,25 +19,18 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"slices"
```
rikatz, line 22: the failures here are happening because probably you also need to bump it into github actions like https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/blob/main/.github/workflows/pullrequests.yaml#L17 and https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/blob/main/.github/workflows/pullrequests.yaml#L30 and https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/blob/main/.github/workflows/pullrequests.yaml#L41

rikatz:
/lgtm
(I know this lgtm will be gone once you fix the CI, but I can re-add it). Will leave final approval for @schrej 

I think as a next thing (I can help with it!) I would like to add an e2e test here, that compiles the container and keeps adding globalincluster, localincluster, ipaddress, etc so we can do a request/response and are sure that what we expect to happen really happens :D 

Thank you very very very much @p-strusiewiczsurmacki-mobica and @schrej this is great!!!

rikatz:
@schrej anything else I'm missing?

PR #214, opened 2023-12-07
⚠️ align prefix of deployments with image name

schrej:
Replaces the `caip-in-cluster` prefix with `capi-ipam-in-cluster` to be in line with the future image name. It was pointed out that caip just looks like a misspelling of capi.

I think this will lead to issues when upgrading existing deployments with clusterctl and would require to manually uninstall the provider first.

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -1,18 +1,18 @@
 # Adds namespace to all resources.
-namespace: caip-in-cluster-system
+namespace: capi-ipam-in-cluster-system
 
 # Value of this field is prepended to the
 # names of all resources, e.g. a deployment named
 # "wordpress" becomes "alices-wordpress".
 # Note that it should also match with the prefix (text before '-') of the namespace
 # field above.
-namePrefix: caip-in-cluster-
+namePrefix: capi-ipam-in-cluster-
 
 # Labels to add to all resources and selectors.
 commonLabels:
   cluster.x-k8s.io/provider: "ipam-in-cluster"
 
-bases:
+resources:
 - ../crd
 - ../rbac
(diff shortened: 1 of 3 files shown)


rikatz:
/lgtm
/hold 

Unhold as you wish :) 

Gave a headsup on it for my colleagues as well! Thanks!

schrej:
Thanks. Then we'll just have to remember putting this into release notes. Should only require a manual removal of the old resources. Since CRDs can stay, no data should be lost.

/unhold

agtogna:
> [...] Should only require a manual removal of the old resources. Since CRDs can stay, no data should be lost.

I desume it can't be upgraded via `clusterctl`. What is the correct way to upgrade then?
Could we completely remove the "old" `caip-in-cluster-system` namespace and then upgrade the IPAM component via clusterctl? e.g.
```shell
kubectl delete namespace caip-in-cluster-system
clusterctl upgrade apply --ipam incluster:v0.1.0-rc.0
```

schrej:
Yeah, that should work. I haven't tested it though.

agtogna:
the correct way to do it is:
```shell
kubectl delete namespace caip-in-cluster-system
clusterctl init --ipam incluster:v0.1.0 --config clusterctl-IPAM.config.yaml
```
where clusterctl-IPAM.config.yaml is:
```yaml
---
providers:
  - name: incluster
    url: https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/releases/latest/ipam-components.yaml
    type: IPAMProvider
```

schrej:
`clusterctl` v1.7 should no longer require the extra configuraiton.

PR #253, opened 2024-05-11
🐛 Add reconciler watch for when pool capacity becomes available

### LINKED ISSUE #179: Claims that are stuck waiting for an address when a pool has no free addresses should be re-reconciled when addresses become available
I noticed an issue recently when creating multiple clusters with the same pool and forgetting that my pool size was too small for the number of machines I created. The observed behavior was that when the pool runs out of IPs, the cluster will never come up because it's waiting for a node to obtain an IP address. Even when deleting an unused cluster, this issue does not resolve itself. I have to manually delete the claim and its owner resources in order to get the deployment to progress.

# Reproduction

1. Create a pool with N addresses, and reserve all of them by creating N claims
2. Attempt to create another claim
3. This claim should be forever waiting for an IP address
4. Delete some claims to free up capacity in this pool
5. Observe that the N+1 claim will still be unable to reserve an IP address.

# Proposed Solution

We should add a watch for when IPAddressClaims are deleted to trigger reconciles on other IPAddressClaims.

flawedmatrix:
This PR adds a watch to the IPAddressClaim reconciler to allow it to progress claims that are still waiting for an available IP address when free addresses become available in the pool. It accomplishes this by adding a predicate only in the case where Free addresses in the Pool status are increased from 0.

Fixes #179


internal/controllers/ipaddressclaim_test.go
```diff
@@ -1057,6 +1057,249 @@ var _ = Describe("IPAddressClaimReconciler", func() {
...
+				Expect(k8sClient.Create(context.Background(), &claim)).To(Succeed())
+
+				addresses := ipamv1.IPAddressList{}
+				Consistently(ObjectList(&addresses)).
+					WithTimeout(10 * time.Second).WithPolling(100 * time.Millisecond).Should(
+					HaveField("Items", HaveLen(1)))
+
+				// Sleep 5 seconds to give the reconciler time to stop retrying
+				// on the newly created claim.
+				time.Sleep(5 * time.Second)
```
schrej, line 1162: not ideal, but I guess there is no other way

internal/controllers/ipaddressclaim.go
```diff
@@ -256,3 +262,29 @@ func resourceTransitionedToUnpaused() predicate.Predicate {
 		},
 	}
 }
+
+func poolStatus(o client.Object) *v1alpha2.InClusterIPPoolStatusIPAddresses {
+	if ipPool, ok := o.(*v1alpha2.InClusterIPPool); ok {
```
schrej, line 267: the linter is right, a type switch would be cleaner here.
https://gobyexample.com/switch (last example)

internal/controllers/ipaddressclaim_test.go
flawedmatrix: Yeah I didn't want to add this either but it wouldn't reproduce the issue otherwise.

internal/controllers/ipaddressclaim.go
flawedmatrix: Thanks for pointing it out. I've got it fixed.

schrej:
Will need to be rebased after #250 is in for the tests to pass.

PR #260, opened 2024-05-27
🌱 Introduce Metric as infospec on Gateways

65278:
The api holds a Gateway string. We're using this in our `cluster-api-provider-proxmox` to create a default route (via netplan in bootstrap). We do happen to sometimes have multiple default gateways. In this case it'd be advantageous to transport a  metric with a default gateway. For this reason, I've added such a thing.

Routes in linux (and win32) fit into unsigned 32 bit integers, as such, neither validation nor testing is required. I've also made it a pointer so it can be nil.

A more complete implementation would require changing the API. I.e.: a network can have multiple gateways, with each one having an info spec, and the info spec being of considerable size: https://man7.org/linux/man-pages/man8/ip-route.8.html#SYNOPSIS .

An alternative would be to remove the gateway field altogether (it serves no purpose over exclusions, really) and instead create a fully featured gateway resource which has an ipam resource object as a parent, but I much prefer this minimal change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,20 @@ spec:
   allocateReservedIPAddresses: true
 ```
 
+To support using the `gateway` field as a default route, a metric can be added for informational purposes. 
+
+```yaml
+apiVersion: ipam.cluster.x-k8s.io/v1alpha2
+kind: InClusterIPPool
+metadata:
+  name: inclusterippool-sample
+spec:
+  addresses:
+    - 10.0.0.0/24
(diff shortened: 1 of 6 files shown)


65278:
I now understand that I have to request a change to the reference api in cluster-api. I'll put this on hold indefinitely.

schrej:
Please open an issue on the cluster-api repository first, before starting with the PR directly.

PR #275, opened 2024-08-21
✨ add support for the claim.spec.clusterName property

### LINKED ISSUE #258: Add support for IPAddressClaim spec.clusterName once CAPI 1.8 is released
IPAddressClaim now has a `spec.clusterName` field, which should be preferred over the `cluster.x-k8s.io/cluster-name` annotation.

https://github.com/kubernetes-sigs/cluster-api/pull/10182

schrej:
Adds support for the new `spec.clusterName` of `IPAddressClaims` that was introduced with CAPI 1.8. It is respected when checking whether the referenced Cluster is paused. The label is also still supported.

I've added an index for claims that gets filled with either the property or the label's value to make it easier to react to cluster unpausing.

Also contains a few dependency upgrades, including CAPI 1.8.1

Fixes #258


internal/controllers/ipaddressclaim_test.go
```diff
@@ -1599,7 +1599,7 @@ var _ = Describe("IPAddressClaimReconciler", func() {
 		})
 	})
 
-	Context("When the cluster is spec.paused true and the ipaddressclaim has the cluster-name label", func() {
+	Context("When the cluster is spec.paused true and the ipaddressclaim spec.clusterName set", func() {
```
mcbenjemaa, line 1602:
```suggestion
	Context("When the ipaddressclaim spec.clusterName reference a paused cluster", func() {
```

internal/controllers/ipaddressclaim_test.go
```diff
@@ -1861,20 +1888,66 @@ var _ = Describe("IPAddressClaimReconciler", func() {
...
 				Expect(k8sClient.Update(context.Background(), &cluster)).To(Succeed())
 
-				Eventually(ObjectList(&addresses)).
+				Eventually(ObjectList(&addresses, client.InNamespace(namespace))).
 					WithTimeout(5 * time.Second).WithPolling(100 * time.Millisecond).Should(
 					HaveField("Items", HaveLen(1)))
 			})
 		})
 
+		Context("When the cluster is referenced using the label and can be retrieved", func() {
```
mcbenjemaa, line 1905:
```suggestion
		Context("When the ipaddressclaim reference a paused cluster via cluster label", func() {
```

pkg/ipamutil/reconciler.go
```diff
@@ -273,12 +294,7 @@ func (r *ClaimReconciler) reconcileDelete(ctx context.Context, claim *ipamv1.IPA
 func (r *ClaimReconciler) clusterToIPClaims(_ context.Context, a client.Object) []reconcile.Request {
 	requests := []reconcile.Request{}
 	claims := &ipamv1.IPAddressClaimList{}
-	err := r.Client.List(context.Background(), claims, client.MatchingLabels(
-		map[string]string{
-			clusterv1.ClusterNameLabel: a.GetName(),
-		},
-	))
-	if err != nil {
+	if err := r.Client.List(context.Background(), claims, client.MatchingFields{"clusterName": a.GetName()}); err != nil {
```
mcbenjemaa, line 297: does this also index the ClusterNameLabel ?

pkg/ipamutil/reconciler.go
```diff
@@ -81,6 +81,22 @@ func (r *ClaimReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager
 		return fmt.Errorf("error setting the manager: Adapter is nil")
 	}
 
+	if err := mgr.GetFieldIndexer().IndexField(ctx, &ipamv1.IPAddressClaim{}, "clusterName", func(object client.Object) []string {
```
mcbenjemaa, line 84: Maybe we create a function for this.

pkg/ipamutil/reconciler.go
schrej: Yes, it prefers the field but indexes both.

PR #316, opened 2025-02-14
:seedling: add webhook flags

defo89:

For local development and some cases where webhooks cannot be used, adds an option via env variable to disable them.

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,8 +21,6 @@ import (
 	"flag"
 	"os"
 
-	"sigs.k8s.io/controller-runtime/pkg/webhook"
-
 	//+kubebuilder:scaffold:imports
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -36,6 +34,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
 	"sigs.k8s.io/cluster-api-ipam-provider-in-cluster/api/v1alpha1"
 	"sigs.k8s.io/cluster-api-ipam-provider-in-cluster/api/v1alpha2"
--- a/main.go
+++ b/main.go
@@ -21,6 +21,8 @@ import (
 	"flag"
 	"os"
 
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+
 	//+kubebuilder:scaffold:imports
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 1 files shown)


hardikdr (approved):
This is certainly useful, thanks for the PR.

schrej:
I'm not sure how useful this is. If you want to disable webhooks, you'll need to delete the configuration resources from the API server. If you only disable them within the provider the API server won't be able to reach the webhooks and fail every time.
If the webhooks aren't configured on the API server, they won't be called, even when they are still served by the provider.

defo89:
> I'm not sure how useful this is. If you want to disable webhooks, you'll need to delete the configuration resources from the API server. If you only disable them within the provider the API server won't be able to reach the webhooks and fail every time. If the webhooks aren't configured on the API server, they won't be called, even when they are still served by the provider.

Sure, whoever is doing this needs to make sure to omit deploying the resources and not to mount the respective secret in Deployment.

schrej:
What's the benefit of disabling them though? Do other providers also support this?

defo89:
> What's the benefit of disabling them in the provider though? Is disabling the webhook configuration not sufficient?

Yes, both controller config and not deploying webhook resources are needed to disabling webhooks. Not disabling them in the provider leads to manager shutting down due to missing cert:
```
2025-02-25T10:32:02.812320520Z I0225 10:32:02.812258       1 internal.go:541] "Stopping and waiting for webhooks"
2025-02-25T10:32:02.812324141Z E0225 10:32:02.812259       1 kind.go:76] "failed to get informer from cache" err="Timeout: failed waiting for *v1alpha2.GlobalInClusterIPPool Informer to sync" logger="controller-runtime.source.EventHandler"
2025-02-25T10:32:02.812349844Z I0225 10:32:02.812295       1 internal.go:544] "Stopping and waiting for HTTP servers"
2025-02-25T10:32:02.812368849Z I0225 10:32:02.812334       1 server.go:254] "Shutting down metrics server with timeout of 1 minute" logger="controller-runtime.metrics"
2025-02-25T10:32:02.812372048Z E0225 10:32:02.812338       1 kind.go:76] "failed to get informer from cache" err="Timeout: failed waiting for *v1alpha2.InClusterIPPool Informer to sync" logger="controller-runtime.source.EventHandler"
2025-02-25T10:32:02.812445698Z I0225 10:32:02.812391       1 server.go:68] "shutting down server" name="health probe" addr="[::]:8081"
2025-02-25T10:32:02.812504998Z I0225 10:32:02.812457       1 internal.go:548] "Wait completed, proceeding to shutdown the manager"
2025-02-25T10:32:02.812531217Z E0225 10:32:02.812499       1 main.go:160] "problem running manager" err="open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory" logger="setup"
```

>  Do other providers also support this?
I have seen same approach in many controller and in capi providers, too. e.g. 
https://github.com/linode/cluster-api-provider-linode/blob/main/cmd/main.go#L230
https://github.com/k3s-io/cluster-api-k3s/blob/main/bootstrap/main.go#L99

schrej:
Fine, we can add a way to disable webhooks. Instead of some env variable, would you also be fine with adding the Webhook flags that CAPI has and use `--webhook-port=0` to disable them entirely?
A consistent solution for this among all providers would be great. I personally would prefer a flag over a env variable.

```go
	fs.IntVar(&webhookPort, "webhook-port", 9443,
		"Webhook Server port")

	fs.StringVar(&webhookCertDir, "webhook-cert-dir", "/tmp/k8s-webhook-server/serving-certs/",
		"Webhook cert dir.")

	fs.StringVar(&webhookCertName, "webhook-cert-name", "tls.crt",
		"Webhook cert name.")

	fs.StringVar(&webhookKeyName, "webhook-key-name", "tls.key",
		"Webhook key name.")
```

defo89:
Suggested approach sounds good to me. I have updated the PR accordingly.

PR #317, opened 2025-02-19
:bug: fix skipping reconciliation, when a referenced cluster is already deleted

llxp:
Currently there is a bug, that if you delete a cluster, which is referenced in an IPAddressClaim, the IPAddressClaim won't be deleted and in the log, the following message is shown instead:
`IPAddressClaim linked to a cluster that is not found, unable to determine cluster's paused state, skipping reconciliation`
This is due to the function `r.reconcileDelete()` only deleting the ReleaseAddressFinalizer in the local copy of the claim and not in the actual object. The update of the object is missing.

This PR fixes this behaviour by adding the update clause to update the object in k8s/etcd.

### DIFF
--- a/pkg/ipamutil/reconciler.go
+++ b/pkg/ipamutil/reconciler.go
@@ -140,11 +140,12 @@ func (r *ClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			if !claim.ObjectMeta.DeletionTimestamp.IsZero() {
+				patch := client.MergeFrom(claim.DeepCopy())
 				if err := r.reconcileDelete(ctx, claim); err != nil {
 					return ctrl.Result{}, fmt.Errorf("reconcile delete: %w", err)
 				}
-				// we'll need to explicitly update the claim here since we haven't set up a patch helper yet.
-				if err := r.Client.Update(ctx, claim); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/ipamutil/reconciler.go
```diff
@@ -140,7 +140,14 @@ func (r *ClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			if !claim.ObjectMeta.DeletionTimestamp.IsZero() {
-				return ctrl.Result{}, r.reconcileDelete(ctx, claim)
+				if err := r.reconcileDelete(ctx, claim); err != nil {
+					return ctrl.Result{}, fmt.Errorf("reconcile delete: %w", err)
+				}
+				// we'll need to explicitly update the claim here since we haven't set up a patch helper yet.
+				if err := r.Client.Update(ctx, claim); err != nil {
```
mcbenjemaa, line 147: one remark:
I know that the patch helper is not initialized yet,
can we use patch instead?

mcbenjemaa (approved):
LGTM otherwise

pkg/ipamutil/reconciler.go
llxp: if you mean to replace r.Client.Update with r.Client.Patch, then this seems to work as well. I added a commit with changes for both finalizers. Tests seem to still work as well with this in place.

schrej (approved):
/approve

PR #318, opened 2025-02-24
🐛 upgrade to CAPI v1.9.5 and align c-r with it to unblock importing from other providers

schrej:
Upgrades CAPI to v1.9.5 and downgrades controller-runtime to v0.19.6 to align it with the version used by CAPI. This is necessary to facilitate importing packages in other ipam providers when they also depend on other CAPI providers, e.g. for testing.

Fixes #


mcbenjemaa (commented):
Did you tried to pin the dependency in the project?

schrej:
We can manually downgrade using a `replace` in go.mod, but I'd prefer it if it matches. Makes it a lot more user-friendly.

PR #329, opened 2025-03-28
:bug: properly handle deletion of IPAddressClaims in ipamutil.ClaimReconciler

llxp:
This PR is based on the changes from @schrej and fixes a bug, where ReleaseAddress is not getting called during deletion.
Additionally it fixes an error regarding to error handling:
`err := errors.New("pool not found")`
This hides the error of FetchPool(). Instead it should be wrapped/enhanced by the new error.
And changes some behaviour regarding the finalizer patching. The defer is moved down and the finalizer patching is now done explicitly instead of implicitly by the defer.

### DIFF
--- a/pkg/ipamutil/reconciler.go
+++ b/pkg/ipamutil/reconciler.go
@@ -176,7 +176,9 @@ func (r *ClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
 		}
 	}()
 
-	controllerutil.AddFinalizer(claim, ReleaseAddressFinalizer)
+	if controllerutil.AddFinalizer(claim, ReleaseAddressFinalizer) {
+		return ctrl.Result{}, nil
+	}
 
 	var res *reconcile.Result
 	var pool client.Object
--- a/pkg/ipamutil/reconciler.go
+++ b/pkg/ipamutil/reconciler.go
@@ -187,7 +187,7 @@ func (r *ClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
 	handler := r.Adapter.ClaimHandlerFor(r.Client, claim)
 	if pool, res, err = handler.FetchPool(ctx); err != nil || res != nil {
 		if apierrors.IsNotFound(err) {
-			err := errors.New("pool not found")
+			err := fmt.Errorf("pool not found: %w", err)
 			log.Error(err, "the referenced pool could not be found")
 			if !claim.ObjectMeta.DeletionTimestamp.IsZero() {
 				return r.reconcileDelete(ctx, claim, handler)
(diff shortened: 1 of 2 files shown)


schrej:
I've closed the duplicate, but will keep the hold for further internal testing in conjunction with our Infoblox provider.

pkg/ipamutil/reconciler.go
```diff
@@ -170,14 +170,10 @@ func (r *ClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
...
+	if controllerutil.AddFinalizer(claim, ReleaseAddressFinalizer) {
 		if err := patchHelper.Patch(ctx, claim); err != nil {
-			reterr = kerrors.NewAggregate([]error{reterr, err})
+			return ctrl.Result{}, err
 		}
-	}()
-
-	if controllerutil.AddFinalizer(claim, ReleaseAddressFinalizer) {
-		return ctrl.Result{}, nil
 	}
```
schrej, line 177: Why is this approach better?

pkg/ipamutil/reconciler.go
```diff
@@ -187,7 +183,7 @@ func (r *ClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
 	handler := r.Adapter.ClaimHandlerFor(r.Client, claim)
 	if pool, res, err = handler.FetchPool(ctx); err != nil || res != nil {
 		if apierrors.IsNotFound(err) {
-			err := errors.New("pool not found")
+			err := fmt.Errorf("pool not found in in-cluster-provider: %w", err)
```
schrej, line 186: the `ipamutil.Reconciler` is supposed to be generic. We shouldn't mention the in-cluster provider here.

schrej (changes requested):

pkg/ipamutil/reconciler.go
llxp: Because it is more explicit code flow instead of the implicit defer, less side effects and better control, meaning it is easier to understand and to read imo, which I think makes it also easier to debug as a result. It caused some problems with reconcileDelete function with double patching/deletion. I probably also would completely remove the defer, but I wasn't sure what potential side effects that could bring for the rest of the code, especially everything inside `controllerutil.CreateOrPatch` function. I only was sure about the changes up to the line where I moved it, without changing its current behaviour.

pkg/ipamutil/reconciler.go
llxp: changed that.

pkg/ipamutil/reconciler.go
schrej: The defer isn't implicit. I do agree that it's a bit harder to understand than a linear implementation. The main advantage of using defer here is that we do not need to take care of manually patching when we change something. The patch helper will automatically check whether something was changed, and update the resource on the API if necessary. Since we're only deferring once, it should only patch one time.
The CreateOrPatch call is responsible for creating or patching the IPAddress, while the patch helper is concerned with the IPAddressClaim.

PR #338, opened 2025-06-05
:bug: Validate the cache contains the created address before moving on

rikatz:
IPAM controller relies on controller runtime client, which is cached, to list the existing address before allocating a new IP.

When the cache is too slow to reflect the changes, we may end up with an IPAddressClaim reconciliation happening before the previous reconciliation address acknowledge by the cache, which may end with duplicated IP address allocations.

This change enforces that the Reconcile function waits for the cached client to be aware of the IPAddress before allowing the reconciliation to move on to the next reconciliation request

Fixes #


pkg/ipamutil/reconciler.go
```diff
@@ -245,6 +247,26 @@ func (r *ClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
...
+	// to the next reconciliation request.
+	// This is required, otherwise if the next "EnsureAddress" is called too fast, or the controller-runtime cached client informer
+	// is too slow, we may end up not seeing that the IP Address is being used and can provide duplicated IPs.
+	err = wait.PollUntilContextTimeout(ctx, 5*time.Millisecond, 5*time.Second, true, func(ctx context.Context) (bool, error) {
+		key := client.ObjectKeyFromObject(&address)
+		if err := r.Client.Get(ctx, key, &ipamv1.IPAddress{}); err != nil {
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			}
+			return false, err
```
schrej, line 260: nit: we could use `return false, client.IgnoreNotFound(err)` here instead

schrej (approved):
good find, just one nit.

pkg/ipamutil/reconciler.go
rikatz: fixed and rebased already :) 

lmk if this is the request! thanks!

rikatz:
> good find, just one nit.

yeah, don't ask me how I was able to find it (spoiler alert: a very slow environment :P )

PR #383, opened 2026-02-24
✨: Add conditions on reconciliation of IPClaim for v1beta2

mcbenjemaa:
###  Summary                                                                                                                                                                                                                                 
         
  - Add v1beta2 condition reporting on IPAddressClaim during reconciliation, setting the Ready condition with appropriate reasons: AddressAllocated (success), PoolNotReady, PoolExhausted, and AllocationFailed                          
  - Introduce PoolExhaustedError type in pkg/ipamutil so claim handlers can signal pool exhaustion to the reconciler                                                                                                                      
  - Export ErrPoolExhausted sentinel error in internal/poolutil for consistent error checking
  - Add integration tests verifying the Ready condition is set correctly in each scenario


internal/controllers/ipaddressclaim.go
```diff
@@ -168,7 +169,7 @@ func (h *IPAddressClaimHandler) FetchPool(ctx context.Context) (client.Object, *
 	if h.claim.Spec.PoolRef.Kind == inClusterIPPoolKind {
 		icippool := &v1alpha2.InClusterIPPool{}
 		if err := h.Get(ctx, types.NamespacedName{Namespace: h.claim.Namespace, Name: h.claim.Spec.PoolRef.Name}, icippool); err != nil {
-			return nil, nil, errors.Wrap(err, "failed to fetch pool")
+			return nil, nil, pkgerrors.Wrap(err, "failed to fetch pool")
```
schrej, line 172:
```suggestion
			return nil, nil, fmt.Errorf("failed to fetch pool: %w", err)
```
I didn't even know `pkg/errors` includes a stack when creating errors. Not sure if it's worth keeping that. Iirc the reason for using that lib was that `fmt.Errorf` didn't exist at the time.

internal/controllers/ipaddressclaim.go
```diff
@@ -180,7 +181,7 @@ func (h *IPAddressClaimHandler) FetchPool(ctx context.Context) (client.Object, *
 	}
 
 	if h.pool == nil {
-		err := errors.New("pool not found")
+		err := pkgerrors.New("pool not found")
```
schrej, line 184:
```suggestion
		err := errors.New("pool not found")
```

pkg/ipamutil/reconciler.go
```diff
@@ -2,11 +2,13 @@ package ipamutil
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
-	"github.com/pkg/errors"
+	pkgerrors "github.com/pkg/errors"
```
schrej, line 9: same as above, should we use stdlib (errors, fmt) instead?

schrej (changes requested):
I don't mind adding this to the `ipamutil.ClaimReconciler`, but we should make it configurable since users might be setting conditions on their own already (I am [here](https://github.com/telekom/cluster-api-ipam-provider-infoblox/blob/main/internal/controllers/ipaddressclaim.go#L132) for example).

(and sorry for the very late review on this)

mcbenjemaa:
Superseded by #421, which rebases this work on current main and addresses the review feedback about making the condition-setting configurable.

PR #384, opened 2026-02-25
✨: Add spec.addressReuseGracePeriodSeconds in IPPool to control the time before releasing the IPs

### LINKED ISSUE #332: Released IP should be used again after some delay.
**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

On K8s upgrades, the new and old nodes are part of a cluster until the new nodes become ready.
There is a race condition where the IP is the same, which confuses the API server and prevents the old nodes from being cleaned up.

We are sure that this won't happen if we prevent the use of an already released IP immediately. 
Using released IPs will be possible after some Delays like 5m 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

/kind feature

mcbenjemaa:
### Summary                                                                                                                                                                                                                               
                                                                                                                                                                                      
  Adds an addressReuseGracePeriodSeconds field to InClusterIPPoolSpec that delays IP address release after a claim is deleted, preventing immediate reuse. This avoids IP conflicts during rolling upgrades where old and new nodes
  briefly coexist, and prevents ARP cache issues on network switches.

  How it works: When a claim is deleted, ReleaseAddress() checks the pool's grace period and returns a RequeueAfter result, keeping the IPAddress object alive. Since ListAddressesInUse() naturally includes existing IPAddress objects,
  the IP remains reserved without any changes to the allocation logic.

  - Add AddressReuseGracePeriodSeconds *int32 to InClusterIPPoolSpec (follows CAPI v1.11 *int32 + "Seconds" suffix convention)
  - Fix reconcileDelete to honor non-nil Result from ReleaseAddress() (was silently ignored)
  - Implement grace period logic using claim.DeletionTimestamp as the release timer
  - Skip grace period when pool is not found (cluster teardown)
  - Add webhook validation, integration tests, and CRD updates

  Test plan

  - Webhook test: negative addressReuseGracePeriodSeconds rejected
  - Webhook test: valid positive value accepted
  - Integration test: IPAddress deletion delayed until grace period elapses
  - Integration test: IP not reused during grace period
  - Integration test: grace period skipped when pool is deleted (cluster teardown)
  
  
    Fixes #332


internal/controllers/ipaddressclaim_test.go
```diff
@@ -2113,6 +2113,133 @@ var _ = Describe("IPAddressClaimReconciler", func() {
...
+				}
+				Consistently(Get(&address)).
+					WithTimeout(1 * time.Second).WithPolling(200 * time.Millisecond).Should(Succeed())
+
+				// After grace period, IPAddress should be deleted
+				Eventually(Get(&address)).
+					WithTimeout(5 * time.Second).WithPolling(500 * time.Millisecond).Should(Not(Succeed()))
+			})
+
+			It("should not reuse an IP during the grace period", func() {
```
schrej, line 2171: Does it make sense to keep this test? Since we're already testing allocation logic, just blocking the IP should suffice, which is tested above.

schrej (approved):

mcbenjemaa:
Closed in favor of https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/pull/387

PR #385, opened 2026-02-26
✨ variable-length prefix allocation for IPv6 pools

### LINKED ISSUE #381: Variable-length prefix allocation for IPv6 pools
Here I am willing to propose to introduce `InClusterPrefixPool` and `GlobalInClusterPrefixPool` types for allocating network prefixes from larger aggregates, using the existing `IPAddressClaim` -> `IPAddress` contract.

These new types follow the same pattern as `InClusterIPPool` and `GlobalInClusterIPPool`. Where the existing pool types allocate individual addresses, the prefix pool types allocate sub-prefixes of a configured length.

## Why this matters

Many network environments require allocating a prefix per consumer (for example, per node or per interface context), not just a single host address. This is especially common in IPv6 deployments where per-node `/64` assignment is standard practice:

- RFC 4291, section 2.5.1 (64-bit interface identifier boundary): https://datatracker.ietf.org/doc/html/rfc4291#section-2.5.1
- RFC 4862, section 5.5.3 (SLAAC requires `/64` prefixes): https://datatracker.ietf.org/doc/html/rfc4862#section-5.5.3
- RFC 7421 (why `/64` is the operational boundary): https://datatracker.ietf.org/doc/html/rfc7421

## How it works

`InClusterPrefixPool` works by analogy to `InClusterIPPool`:

- Both have the same pair of resources: namespaced and cluster-scoped (global).
- Both use the same claim/address contract: `IPAddressClaim` references the pool, and an `IPAddress` is created with the allocation result.
- Both share the same spec shape: `addresses`, `prefix`, `gateway`, `excludedAddresses`, `allocateReservedIPAddresses`.
- Both track status the same way: `total`, `used`, `free`, `outOfRange`.

The difference is in what gets allocated:

| | `InClusterIPPool` | `InClusterPrefixPool` |
|---|---|---|
| Allocation unit | Individual IP address | Network prefix (e.g. `/64`) |
| `IPAddress.spec.address` | Host address | Prefix base address |
| `IPAddress.spec.prefix` | Network prefix from pool | Allocated prefix length |
| Status counts | Individual addresses | Prefix units |
| `spec.addresses` format | IPs, ranges, CIDRs | CIDRs only |

## What to expect from this feature

1. Allocation is deterministic (same candidate order, first valid free prefix wins).
2. Already allocated prefixes never overlap with newly allocated prefixes.
3. Expanding `spec.addresses` increases capacity immediately, without restart.
4. Existing `IPAddressClaim`/`IPAddress` contract is reused unchanged.
5. `InClusterIPPool` behavior is completely unaffected.

## Initial scope

The initial implementation may targets IPv6 `/64` allocation to keep things simple:

1. Only IPv6 addresses are accepted in `spec.addresses`.
2. Prefix length is hardcoded to `/64`.
3. `spec.addresses` entries must be CIDRs with prefix bits <= 64.

This covers the most common practical use case. Support for other prefix lengths and IPv4 can be added later without breaking changes.

## Expected CRD/API changes

Two new CRDs, by analogy to the existing pair:

- `InClusterPrefixPool` (namespaced) — like `InClusterIPPool`
- `GlobalInClusterPrefixPool` (cluster-scoped) — like `GlobalInClusterIPPool`

Both share `InClusterPrefixPoolSpec`:

```go
type InClusterPrefixPoolSpec struct {
    Prefixes []string `json:"prefixes"`

    // +kubebuilder:default:=64
    // +kubebuilder:validation:Minimum=1
    // +kubebuilder:validation:Maximum=128
    // +optional
    AllocationPrefixLength int `json:"allocationPrefixLength,omitempty"`

    // +optional
    Gateway string `json:"gateway,omitempty"`

    // +optional
    ExcludedPrefixes []string `json:"excludedPrefixes,omitempty"`
}
```

### Example

```yaml
apiVersion: ipam.cluster.x-k8s.io/v1alpha2
kind: InClusterPrefixPool
metadata:
  name: ipv6-prefix-pool
  namespace: default
spec:
  prefixes:
    - 2001:db8:100::/48
    - 2001:db8:100:1::/64
    - 2001:db8:100:2::/64
  allocationPrefixLength: 64
  excludedPrefixes:
    - 2001:db8:100:ffff::/64
  gateway: 2001:db8:100::1
```

A claim referencing this pool:

```yaml
apiVersion: ipam.cluster.x-k8s.io/v1beta2
kind: IPAddressClaim
metadata:
  name: node-prefix-claim
  namespace: default
spec:
  poolRef:
    apiGroup: ipam.cluster.x-k8s.io
    kind: InClusterPrefixPool
    name: ipv6-prefix-pool
```

The resulting `IPAddress`:

```yaml
apiVersion: ipam.cluster.x-k8s.io/v1beta2
kind: IPAddress
metadata:
  name: node-prefix-claim
  namespace: default
spec:
  address: "2001:db8:100::"
  prefix: 64
  gateway: "2001:db8::1"
  poolRef:
    apiGroup: ipam.cluster.x-k8s.io
    kind: InClusterPrefixPool
    name: ipv6-prefix-pool
```

## Expected runtime behavior details

1. Existing `InClusterIPPool` and `GlobalInClusterIPPool` are completely unaffected.
2. Prefix pools allocate only non-overlapping prefixes.
3. Allocation works reliably from larger aggregates.
4. Pool expansion increases available allocations without restart.
5. Status counts correctly reflect prefix units.
6. Webhook validates addresses are CIDRs, within the aggregate subnet, and IPv6 (for initial scope).

## Alternative approaches

- **`prefixLength` field on `InClusterIPPoolSpec`** — Adds a mode switch to the existing type where validation rules and allocation behavior depend on whether the field is set. A separate type keeps each resource focused on one allocation behavior.

/kind feature

pasteley:

- Adds new pool types for prefix allocation:
  - `InClusterPrefixPool` (namespaced)
  - `GlobalInClusterPrefixPool` (cluster-scoped)
- Implements deterministic, non-overlapping allocation of child IPv6 prefixes.
- Supports configurable allocation size (`allocationPrefixLength`, default `/64`) and excluded ranges/prefixes.
- Adds webhook validation/defaulting to enforce safe updates and consistent behavior.
- Extends controllers to reconcile prefix pools and report capacity/usage status.
- Wires all generated platform pieces (CRDs, RBAC, webhook manifests, manager setup) and adds test coverage across allocator, webhooks, and controllers.

Closes: #381


api/v1alpha2/inclusterprefixpool_conversion.go
```diff
@@ -0,0 +1,29 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha2
+
+// Hub marks InClusterPrefixPool as a conversion hub.
+func (*InClusterPrefixPool) Hub() {}
```
mcbenjemaa: We don't need a conversion hub for now.

README.md
```diff
@@ -77,6 +82,35 @@ spec:
   allocateReservedIPAddresses: true
 ```
 
+### Prefix Pools
+
+For prefix-per-consumer allocation (for example `/64` per claim), use `InClusterPrefixPool` or `GlobalInClusterPrefixPool`.
+
+> [!NOTE]
+> Crrently only IPv6 prefixes supported.
```
mcbenjemaa:
```suggestion
> Currently only IPv6 prefixes supported.
```

api/v1alpha2/inclusterprefixpool_conversion.go
pasteley: Removed

pasteley:
@schrej Jakob, hello, please take a look

internal/controllers/ipaddressclaim.go
schrej: What do you think about moving all of the claim handling into a `controllers/ipaddressclaim` package and having the handlers in separate files, one `address_handler.go` and one `prefix_handler.go` or something like that?

internal/poolutil/pool.go
```diff
@@ -60,15 +92,10 @@ func ListAddressesInUse(ctx context.Context, c client.Reader, namespace string,
...
-		gv, _ := schema.ParseGroupVersion(a.APIVersion)
-		if gv.Group != "ipam.cluster.x-k8s.io" {
-			continue
-		}
-		addr = append(addr, a)
+	if err != nil {
+		return nil, err
 	}
-	return addr, err
+	return addresses.Items, nil
```
schrej, line 65: Reasoning here is that we don't need to filter since only valid addresses will be indexed?

internal/poolutil/pool.go
schrej, line 1: Can we move all of the prefix stuff to a separate file?

internal/controllers/ipaddressclaim_prefix.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		freePrefix, err := poolutil.FindFreePrefix(poolutil.NewPrefixPoolConfig(poolSpec), inUseIPSet)
+		if err != nil {
+			return nil, fmt.Errorf("failed to find free prefix: %w", err)
+		}
+
+		address.Spec.Address = freePrefix.Addr().String()
+		address.Spec.Gateway = poolSpec.Gateway
+		address.Spec.Prefix = ptr.To(int32(freePrefix.Bits())) //nolint:gosec // Bits() is [0,128] for valid prefixes
```
schrej, line 134: Does lead to different semantics of the Prefix field depending on whether you're assigning an address (size of the pool the address is from) or a pool (size of the assigned pool)?

Since I'm not too familiar with the details of v6, is this even relevant (for routing?) when assigning prefixes?

internal/poolutil/pool.go
```diff
@@ -219,3 +246,288 @@ func AddressStrParses(addressStr string) bool {
...
+	sort.Slice(sorted, func(i, j int) bool {
+		if cmp := sorted[i].Addr().Compare(sorted[j].Addr()); cmp != 0 {
+			return cmp < 0
+		}
+		return sorted[i].Bits() < sorted[j].Bits()
+	})
+	return sorted
+}
+
+func findFreePrefixInAggregate(aggregate netip.Prefix, targetBits int, blockedIPSet *netipx.IPSet) (netip.Prefix, bool) {
```
schrej: can you add comments to the if statements in this function to indicate what they are checking for semantically? Would make it easier to follow what's happening. Might be my limited networking knowledge, but It's not directly obvious imo.

internal/poolutil/pool.go
```diff
@@ -219,3 +246,288 @@ func AddressStrParses(addressStr string) bool {
...
+		return netip.Prefix{}, false
+	}
+
+	if prefix, found := findFreePrefixInAggregate(left, targetBits, blockedIPSet); found {
+		return prefix, true
+	}
+	return findFreePrefixInAggregate(right, targetBits, blockedIPSet)
+}
+
+func countAllocatablePrefixesInAggregate(aggregate netip.Prefix, targetBits int, blockedIPSet *netipx.IPSet) int {
```
schrej: A comment that explains how this works would also be nice, it's a bit hard to follow. Elegant though!

internal/poolutil/pool_test.go
```diff
@@ -611,6 +613,181 @@ var _ = DescribeTable("IPSetCount", func(expectedCount int, addresses ...string)
 	Entry("ipv6 CIDR", 4, "fe80::1/126"),
 )
 
+var _ = Describe("Prefix allocation helpers", func() {
+	It("allocates prefixes deterministically with merged aggregates", func() {
```
schrej, line 617: nit: where is the 'merged' part here? With my limited knowledge I'd assume this would allocate the same prefix for either of those pool prefixes. Should we maybe have a more complex setup here to verify that it properly merges prefixes for allocation?

README.md
```diff
@@ -19,7 +19,12 @@ This provider comes with clusterctl support and can be installed using `clusterc
 
 ## Usage
 
-This provider comes with two resources to specify pools from which addresses can be allocated: the `InClusterIPPool` and the `GlobalInClusterIPPool`. As the names suggest, the former is namespaced, the latter is cluster-wide. Otherwise they are identical. The following examples will all use the `InClusterIPPool`, but all examples work with the `GlobalInClusterIPPool` as well.
+This provider comes with two resource pairs:
+
+- Address allocation: `InClusterIPPool` (namespaced) and `GlobalInClusterIPPool` (cluster-wide)
+- Prefix allocation: `InClusterPrefixPool` (namespaced) and `GlobalInClusterPrefixPool` (cluster-wide)
```
schrej: A note that prefix allocation only supports v6 would be nice

schrej (changes requested):
This looks very good overall, well done!

I have a few comments, mostly around code organization and adding a few comments here and there, but nothing major.

internal/poolutil/pool.go
pasteley: Yeap, exactly, `index.IPPoolRefValue` now scopes results to group

internal/controllers/ipaddressclaim_prefix.go
pasteley: It always indicates the prefix length of the network the address belongs to, but differently: 
- for IP address pools it's the pool's subnet prefix
- for Prefix pools it's the allocated child prefix length

There is not much difference for `inet6` in this regard, you still need mask to specify when assigning to interface or creating a route, it's just that `/64` ordinary minimal routable size

internal/poolutil/pool.go
pasteley: Moved to [internal/poolutil/prefix.go](https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/pull/385/changes#diff-a0cc62a0d5938eb4333799e4a61c808a071f2eb55634d44932f062510555789f)

internal/poolutil/pool.go
pasteley: Done

internal/poolutil/pool.go
pasteley: Done

internal/poolutil/pool_test.go
pasteley: Extended and described scenario in comments

README.md
pasteley: Already present in [note in next section](https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/pull/385/changes#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R90), added here as well

internal/controllers/ipaddressclaim.go
pasteley: That would require some rework, that I would like to avoid in PR that brings new feature.
Let's keep adapter entry point in `internal/controllers/ipaddressclaim.go` and just separate `internal/controllers/ipaddressclaim_address.go` and `internal/controllers/ipaddressclaim_prefix.go` inside current package structure

internal/controllers/ipaddressclaim_prefix.go
schrej: Are we then missing information necessary to configure a route for the gateway? When just handing out IP addresses, the prefix can be used to route any traffic outside of the subnet the address is part of via the gateway. With this implementation you only get the size of the prefix assigned to the machine, but no information about the size of the subnet the prefix was allocated from. Is this an issue?

schrej:
One final [question](https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/pull/385#discussion_r3063652064), otherwise lgtm.

Sorry for the delay, lots to do internally.

internal/controllers/ipaddressclaim_prefix.go
pasteley: Not an issue, this is intentional.` Gateway` field is kept for compatibility with the `IPAddress` spec shape, but it has no effect on allocation logic, unlike IP pools, where the gateway address is excluded from the allocatable range.

Consumer that claims from a `(Global)InClusterPrefixPool` already knows it's getting a prefix, not a single address, so there's should be no ambiguity in how to interpret the resulting `IPAddress` fields. If consumer needs a gateway hint alongside the prefix allocation, they can explicitly set it on the pool (e.g. something like `fe80::1` for a  link-local next-hop), and it will be copied as-is into `IPAddress.Spec.Gateway`. If they don't, which is the common case for prefix delegation, it can be omitted.

In other words, we don't try to derive routing information from the parent prefix or validate the gateway against the pool's address space. It's a purely optional, opaque pass-through field.

schrej (approved):
/approve

PR #387, opened 2026-04-27
✨: Add spec.addressReuseGracePeriodSeconds in IPPool to control the time before releasing the IPs

### LINKED ISSUE #332: Released IP should be used again after some delay.

mcbenjemaa:
### Summary                                                                                                                                                                                                                               
                                                                                                                                                                                      
  Adds an addressReuseGracePeriodSeconds field to InClusterIPPoolSpec that delays IP address release after a claim is deleted, preventing immediate reuse. This avoids IP conflicts during rolling upgrades where old and new nodes
  briefly coexist, and prevents ARP cache issues on network switches.

  How it works: When a claim is deleted, ReleaseAddress() checks the pool's grace period and returns a RequeueAfter result, keeping the IPAddress object alive. Since ListAddressesInUse() naturally includes existing IPAddress objects,
  the IP remains reserved without any changes to the allocation logic.

  - Add AddressReuseGracePeriodSeconds *int32 to InClusterIPPoolSpec (follows CAPI v1.11 *int32 + "Seconds" suffix convention)
  - Fix reconcileDelete to honor non-nil Result from ReleaseAddress() (was silently ignored)
  - Implement grace period logic using claim.DeletionTimestamp as the release timer
  - Skip grace period when pool is not found (cluster teardown)
  - Add webhook validation, integration tests, and CRD updates

  Test plan

  - Webhook test: negative addressReuseGracePeriodSeconds rejected
  - Webhook test: valid positive value accepted
  - Integration test: IPAddress deletion delayed until grace period elapses
  - Integration test: IP not reused during grace period
  - Integration test: grace period skipped when pool is deleted (cluster teardown)
  
  
    Fixes #332


mcbenjemaa:
I have problems force pushing to the working branches.

internal/webhooks/inclusterippool.go
```diff
@@ -232,6 +232,15 @@ func (webhook *InClusterIPPool) validate(_, newPool types.GenericInClusterPool)
 		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "excludedAddresses"), newPool.PoolSpec().ExcludedAddresses, "addresses and excluded addresses are of mixed IP families"))
 	}
 
+	if newPool.PoolSpec().AddressReuseGracePeriodSeconds != nil &&
+		*newPool.PoolSpec().AddressReuseGracePeriodSeconds < 0 {
```
schrej, line 236: can we use `// +kubebuilder:validation:Minimum:=0` instead?

pkg/ipamutil/reconciler.go
```diff
@@ -278,8 +278,11 @@ func (r *ClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ct
 }
 
 func (r *ClaimReconciler) reconcileDelete(ctx context.Context, claim *ipamv1.IPAddressClaim, handler ClaimHandler) (ctrl.Result, error) {
-	if res, err := handler.ReleaseAddress(ctx); err != nil {
-		return unwrapResult(res), fmt.Errorf("release address: %w", err)
+	if res, err := handler.ReleaseAddress(ctx); err != nil || res != nil {
```
schrej: This is a breaking change for consumers of the `ipamutil` package. The current implementation could be classified as a bug, but it still changes behavior: if an implementation returns `&ctrl.Result{}, nil` it would currently continue with deletion. After this change it would block deletion indefinitely.

internal/controllers/ipaddressclaim_address.go
```diff
@@ -158,9 +159,38 @@ func (h *IPAddressClaimHandler) EnsureAddress(ctx context.Context, address *ipam
...
+	}
+
+	gracePeriod := time.Duration(*gracePeriodSeconds) * time.Second
+	elapsed := time.Since(h.claim.DeletionTimestamp.Time)
+	remaining := gracePeriod - elapsed
+	if remaining > 0 {
+		log.Info("Address reuse grace period active, delaying IP address release",
+			"remaining", remaining.Round(time.Second).String(),
+			"gracePeriodSeconds", *gracePeriodSeconds)
+		return &ctrl.Result{RequeueAfter: remaining}, nil
```
schrej, line 191: With the change to `ipamutil` this would block the deletion of the IPAddress (which is the intention). Since we're setting `BlockOwnerDeletion` on the owner reference from the IPAddressClaim, this would also block the deletion of the claim when using [foreground cascading deletion](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#cascading-deletion) (background is the default behavior).
In case any infrastructure provider does so, this might block (infra) Machine deletion. When using a delete-then-create strategy during a rolling upgrade, this would lead to a deadlock. I had to read up on this while writing this comment, and nobody is probably doing this, so maybe we should just ignore it.

schrej (changes requested):
Looks good in general, but some behavior concerns and a suggestion.

internal/webhooks/inclusterippool.go
mcbenjemaa: the validation is already there.

pkg/ipamutil/reconciler.go
mcbenjemaa: I agree, it will make consumers deletion blocked.
I have analysed this and i think i can scope the early-return only to results that actually request requeue `(Requeue || RequeueAfter > 0)`. which eliminates the breakage while still letting the
  grace-period code work.

internal/controllers/ipaddressclaim_address.go
mcbenjemaa: part of this is actually true, after my change in the `reconcileDelete()` to return early only when a requeue is actually requested, this would be a delay rather than a deadlock.
I'm going to document this implication on the field.
I believe after this we are good to go, as the `gracePeriodSeconds` only effective unless explicitly set in the Pool Spec.

mcbenjemaa: sgh! the branch protection is rejecting my push again.

mcbenjemaa: after this PR, I will use my Fork

mcbenjemaa: I need you to remove the branch hook, for the last time, and sorry for this..

pkg/ipamutil/reconciler.go
schrej: Good idea. On the other hand I'd also be fine with a breaking change here, since it was previously not really specified, and since the result is a pointer, I would personally expect it to behave the way you've implemented now (I think I also intended it that way, but didn't implement it properly at the time).

schrej: Considering this basically mimics c-r behavior, doing it this way seems fine.

api/v1alpha2/inclusterippool_types.go
```diff
@@ -47,6 +47,18 @@ type InClusterIPPoolSpec struct {
...
 	ExcludedAddresses []string `json:"excludedAddresses,omitempty"`
+
+	// AddressReuseGracePeriodSeconds is the duration in seconds for which an
+	// IP address should be kept reserved after being released, preventing
+	// immediate reuse. This helps avoid IP conflicts during rolling upgrades
+	// where old and new nodes may briefly coexist, and prevents ARP cache
+	// issues on network switches.
+	// If not set or set to 0, addresses are available for reuse immediately.
+	// During the grace period the IPAddress and claim deletions are held, the downstream consumers using foreground cascading
+	// deletion through the claim will see deletion of upstream resources delayed by up to AddressReuseGracePeriodSeconds..
```
schrej: nit: two dots at the end

internal/controllers/ipaddressclaim_address.go
```diff
@@ -158,9 +159,38 @@ func (h *IPAddressClaimHandler) EnsureAddress(ctx context.Context, address *ipam
...
+
+	gracePeriodSeconds := h.pool.PoolSpec().AddressReuseGracePeriodSeconds
+	// A nil or zero value means no grace period; addresses are available for reuse immediately.
+	if gracePeriodSeconds == nil || *gracePeriodSeconds <= 0 {
+		return nil, nil
+	}
+
+	if h.claim.DeletionTimestamp == nil {
+		return nil, nil
+	}
```
schrej: maybe move this up?

schrej (commented):
One more nit and a minor suggestion.

schrej (approved):
/approve

PR #394, opened 2026-05-13
feat: add --leader-election-namespace flag to manager

videlov:
## Summary

- Adds a new `--leader-election-namespace` CLI flag to the controller manager.
- The flag maps directly to `ctrl.Options.LeaderElectionNamespace`, which is passed to controller-runtime.
- When unset, controller-runtime continues to default to the pod's own namespace — no behaviour change for existing deployments.

## Motivation

Some deployment scenarios require the leader election lock to live in a specific namespace that differs from the pod's namespace (e.g. multi-tenant setups, cross-namespace deployments, or environments where the pod namespace annotation is unavailable). This flag brings the manager in line with other CAPI providers that already expose this knob.

## Test plan

- [ ] Existing unit and e2e test suite passes — no logic was changed, only a new flag is wired through.
- [ ] Manual verification: start the manager with `--leader-elect=true --leader-election-namespace=<ns>` and confirm the `lease` object is created in the specified namespace.

### DIFF
--- a/main.go
+++ b/main.go
@@ -62,10 +62,11 @@ func init() {
 
 func main() {
 	var (
-		enableLeaderElection bool
-		probeAddr            string
-		watchNamespace       string
-		watchFilter          string
+		enableLeaderElection    bool
+		leaderElectionNamespace string
+		probeAddr               string
+		watchNamespace          string
+		watchFilter             string
 
 		managerOptions = flags.ManagerOptions{}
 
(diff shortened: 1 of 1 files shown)


schrej (approved):
Not against adding this, but I'm curious: CAPI does not expose this flag, some providers do. With what setup is it required for providers, but not for core CAPI? Or did I miss it there?

videlov:
@schrej we have a scenario when we use this operator standalone to handle IP pools and addresses allocation, no other CAPI component is involved. since we use specific namespace for different projects we would like to diversity and have the leader election per namespace.

schrej (approved):
/approve

PR #418, opened 2026-09-03
🐛 fix tolerate control-plane taints in the manager Deployment

mlguerrero12:
The in-cluster IPAM provider is the only Cluster API provider whose manager Deployment does not tolerate the control-plane taints. Core Cluster API, the kubeadm bootstrap and control-plane providers, and the vSphere infrastructure provider all carry both tolerations, so this is the only pod-level field that differs between them and this provider.

Without the tolerations the controller cannot be scheduled on a management cluster whose nodes are all control-plane nodes, and it becomes unschedulable on any cluster that temporarily has no other node available. Because this controller sits on the Machine provisioning path, that is a deadlock rather than a degradation: a replacement Machine stays in WaitingForIPAddress because only this controller can bind its IPAddressClaim, while the controller stays Pending waiting for the node that the Machine would have become. Recovering requires manually patching the Deployment or removing the control-plane taint.


mlguerrero12:
@schrej PTAL

schrej:
Wasn't aware, thanks!
/approve
/lgtm

